PR #4, opened 2018-03-15
Improve Maintainers, Links, Info, and add Notes

huyhg:


pkg/apis/application/v1alpha1/application_types.go
```diff
@@ -64,6 +66,37 @@ type ApplicationStatus struct {
...
+	Name string `json:"name,omitempty"`
+
+	// Url could typically be a website address or a mailto: email address.
+	Url string `json:"url,omitempty"`
+}
+
+// Link contains information about an URL to surface documentation, dashboards, etc.
+type Link struct {
+	// Type defines the purpose of the link.
+	Type string `json:"type,omitempty"`
```
deustis: Should there be a valid set of strings for Type? (e.g. "documentation", "support", "marketing", ...)

pkg/apis/application/v1alpha1/application_types.go
```diff
@@ -64,6 +66,37 @@ type ApplicationStatus struct {
...
 }
 
+// Maintainer contains information about an individual or organization that maintains the source code, images, and
+// package for an Application. An Application can have more than one maintainer.
+type Maintainer struct {
+	// Name is the descriptive name of the maintainer.
+	Name string `json:"name,omitempty"`
+
+	// Url could typically be a website address or a mailto: email address.
+	Url string `json:"url,omitempty"`
```
mattfarina: In addition to Name and URL there should be a space for an email address. That would round out the common set of information captured here.

pkg/apis/application/v1alpha1/application_types.go
mattfarina: @deustis On other projects that capture this same kind of information are there cases that capture this form of standardized categorization?

pkg/apis/application/v1alpha1/application_types.go
```diff
@@ -64,6 +66,37 @@ type ApplicationStatus struct {
...
+
+	// Description is human readable content explaining the purpose of the link.
+	Description string `json:"description,omitempty"`
+
+	// Url typically points at a website address.
+	Url string `json:"url,omitempty"`
+}
+
+// InfoItem is a human readable key,value pair containing important information about how to access the Application.
+type InfoItem struct {
```
mattfarina: If the intent is information on how to access the application, rather than general information, should it be `AccessMethod` or something else more specific to the type of information? Or, is this general information rather than access info?

pkg/apis/application/v1alpha1/application_types.go
huyhg: Done. This is consistent with how helm does things right.

My original intention was to have the email address captured as a `mailto:` URL, but supporting both URL and email for a maintainer sounds reasonable.

pkg/apis/application/v1alpha1/application_types.go
huyhg: This is probably not an enum, although we can recommend some convention around the values of this field, for example `application.k8s.io/support`.

Should we document such convention here, or just keep it this way?

pkg/apis/application/v1alpha1/application_types.go
huyhg: We can rename `Info` field to `AccessInfo`. @kow3ns what do you think?

pkg/apis/application/v1alpha1/application_types.go
kow3ns: I think we should actually remove type. I don't see what value it provides.

huyhg: An example of how these fields are used:
```
links:
- type: support
  description: Community support for Wordpress
  url: https://stackoverflow.com/questions/tagged/wordpress/
- type: info
  description: More information about Wordpress
  url: https://wordpress.org/
- type: license
  url: https://wordpress.org/about/license/
```

The `info` link in this particular Wordpress case could be removed and replaced by the maintainer's URL. We can imagine a case where the same maintainer (i.e. company) has many applications. `info` would be application specific and useful.

deustis: @mattfarina, no standards that I'm aware of.

We ([Google Cloud Launcher](https://console.cloud.google.com/launcher)) have experience collecting product metadata from software vendors. We don't try to categorize different link types today, but we do see very common recurring patterns. We could try to pull together some data if that would be helpful.

The main reason to consider categorizing links would be able to give them special treatment in UIs. for example, UIs would typically want to feature the "user guide" more prominently than "tutorials", or "community forums".

I do think we should either remove Type, or restrict it to a specific set of values. Otherwise, as a free-form string, it doesn't add much value.

pkg/apis/application/v1alpha1/application_types.go
deustis: Again, drawing from experience on [Google Cloud Launcher](https://console.cloud.google.com/launcher) --

I'd expect that access-related info (credentials and endpoints) would be most common, but it would also be quite common to highlight other selected values: 
- application configuration
- infrastructure tuning
- scaling parameters

I think it's best to think of this a table of general information: curated by the author of the application to highlight the most salient app-specific values for quick reference.

deustis: On the topic of credentials, we might want to establish a pattern for referencing Secrets. A very common getting started experience for an off-the-shelf app is trying to locate the password for an admin dashboard. An off-the-shelf app might generate a password and store in a Secret (e.g. default behavior of [Helm WordPress](https://github.com/kubernetes/charts/tree/master/stable/wordpress)), but locating that password is non-trivial for a new user (what's a Secret? which Secret? base64 encoding?)

I'd suggest a convention where the value can be a Secret [resource Name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), and perhaps sub-reference into the Secret data

prydonius: @deustis that makes sense to me, an access credential could look something like:
```
- name: username
  value: user
- name: password
  valueFrom:
    secretKeyRef:
       name: wordpress-secret
       key: wordpress-password
```

How can we surface access info? This is typically a Service or Ingress in charts today, and in the former case if the Service is type LoadBalancer, we won't know the IP address/hostname until it's provisioned. Perhaps we should follow the credentials approach here too and reference the Service or Ingress to use to access the application (which is helpful in distinguishing which Service should be used to access the application when there are multiple).

huyhg: My thinking is that the information here should be straight forward to consume. The definition of how it is computed is a spec that can be consumed by a controller who watches the resources and sets the values onto the app CR.

For example:
```
- name: adminUrl
  value: http://1.2.3.4/application/admin
```

The spec for this could look something like the following. However, this spec is not on the application spec--it's defined for the controller only:
```
- name: serviceUrl
  valueFrom:
    serviceRef: wordpress-external-service
- name: adminUrl
  value: http://{{ index .Values 'serviceUrl' }}/application/admin
```

deustis: @huyhg, two questions about your example:
1. What is the value of serviceUrl? Is that the entire contents of the wordpress-external-service object?
2. It seems like the controller would be clobbering the original templated value of adminUrl. Seems like a problem if the value of serviceUrl changes in the future.

deustis: @prydonius, @huyhg, it seems like we've identified three types of resources that would commonly want to be referenced: Secrets, Services, and Ingress. Maybe we should generalize a mechanism to reference any resource/field?:

For example, using JSON Path (where we assume the path resolves to a single value), combined with a base64 decode operation:
```
- name: password
  valueFrom:
    kind: Secret
    name: wordpress-secret
    jsonPath: $.data.wordpress-password
    base64Decode: true
```

deustis: @prydonius, @huyhg, adding in the concept of template expansion:

```
- name: serviceUrl
  valueFrom:
    serviceRef: wordpress-external-service
- name: adminUrl
  templateValue: http://{{ serviceIp }}:{{ servicePort }}/application/admin
  templateData:
    serviceRoot:
      valueFrom:
        kind: Service
        name: wordpress-external-service
        jsonPath: $.spec.clusterIP
    servicePort:
      valueFrom:
        kind: Service
        name: wordpress-external-service
        jsonPath: $.spec.ports[?(@.name==https)].port
```

UIs could directly interpret the value of adminUrl. Or perhaps there would be a standard App controller that would be watch those references and manage the "value" field so that UIs don't have to.

prydonius: I'm not sure there is much gained from generalising this, since there are only 3 (4 incl. potentially ConfigMap) kinds we need to reference at most.

Also, in the above example using JSON Path to get the IP Address doesn't always make sense. Depending on the Service Type, we may want the Cluster IP, Node IP or LoadBalancer IP/Hostname. I think a UI can make this decision by reading the whole Service resource.

I'm not sure about the templating idea, given that it will almost always be a path that should be appended to the Service/Ingress host, what if we just had a `adminPath` key that in the above example would just be `/admin`? Then a UI or tool could find the referenced Service/Ingress, figure out the host from the type, and append `adminPath` onto it.

JSON Path doesn't really seem useful for Secrets either, it would be better to copy the reference stanza from the PodSpec.

konryd: @deustis The problem with runtime template evaluation here is that the same would have to happen for the actual Pods and Deployments that read the data (say, the admin console password). Currently, the secret values are read at pod's startup, so editing the secret manually won't result in the application being updated.

If the application runs some sort of watch on the ConfigMaps (to update the pods on edits) it can also update the Application resource (and its Info field).

deustis: Shall we table this discussion on references and templates? Based on the proposals we've seen in this thread, the current InfoItem object will likely be extensible. I'm happy to open a new issue thread capturing the state of the discussion. It seems like we'll benefit from getting some more feedback on this topic, but we shouldn't block this useful PR in the meantime.

@prydonius, I tend to seek generalization when I encounter a third (or fourth) instance of a pattern, but fully understand that we're reaching here given no established use. 

@konryd, a given application should probably declare (in documentation) whether manually modifying a Secret after installation is supported. Assuming it is supported, then it seems correct to treat the Secret as the source of truth, because Pods are ephemeral and newly scheduled Pods will see the latest value. That would imply that Pods should watch Secrets. But it doesn't imply (in my mind) that Pods should update the Application resource. In particular with Secrets, I would argue that dereferencing the value should only happen in the UI, and only in response to an explicit user action (e.g. click "show password")

konryd: Sounds good. The current proposal is string-to-string, which is a common subset of all the current proposals. Filed #12 to continue this thread.

pkg/apis/application/v1alpha1/application_types.go
huyhg: I'm removing `Type` as it is the only point of contention. We can continue this discussion in another issue/PR.

pkg/apis/application/v1alpha1/application_types.go
huyhg: Sounds great! Let's continue the discussion on #12, unblocking this PR.

pkg/apis/application/v1alpha1/application_types.go
```diff
@@ -38,23 +38,25 @@ type ApplicationSpec struct {
...
+	Info []InfoItem `json:"info,omitempty"`
 
-	// Links is a map of human readable keys to URLs for the Application. This intended to be used to surface additional
-	// documentation, dashboards, etc.
-	Links map[string]string `json:"urls,omitempty"`
+	// Links are a list of descriptive URLs intended to be used to surface additional documentation, dashboards, etc.
+	Links []Link `json:"urls,omitempty"`
+
+	// Notes are a list of human readable snippets intended as a quick start for the users of the Application.
+	Notes []string `json:"notes,omitempty"`
```
konryd: What are the semantics of the list here? Can it be just a single string?

huyhg:
Rebased against the newest master and ran `kubebuilder`.

PR #17, opened 2018-04-06
Adding an icon property to the schema

mattfarina:
Consider the Application is used to display information in a UI, such as dashboard. Having an icon to display alongside the application would be useful.


mattfarina:
@prydonius Since you use icons like this in kubeapps, do you have any feedback on the icon property?

@kow3ns Do we know someone from dashboard or the google cloud console who might have feedback on the icon?

@michelleN Do you know anyone from Azure who might have input on the icon property in the application crd?

prydonius:
In Kubeapps we currently download and process the icons to fit 160x160 (we add padding if they are smaller than this), and this has been okay for the most part with the exception of a few (https://hub.kubeapps.com/).

It might be useful to have different sizes, but I agree with Ken that these will probably need to be mandated. If the sizes are free-form it might be more difficult for UIs to consume. For the purposes of different displays, we may want to indicate whether the image is square or wide. This way I could choose to show just the logo in the square format, and the logo + name in the wide format (e.g cockroachdb could be split for improved readability).

konryd:
I don't think we should go far with mandates for the icon:
 * Application controller won't be able to validate them (unless it ships with image processing libraries)
 * Standardizing the icon pixel sizes across multiple UIs, owned by multiple parties is not a task I sign up for.
 * Scaling the icons can go a long way, within reasonable upper and lower boundaries

OTOH, some light standard is desirable for the field to make sense. As @prydonius notes - free-form shapes will give UIs trouble. How about we require the icon to be square?

@mattfarina: GKE Cloud Console TL here.

README.md
```diff
@@ -79,6 +79,11 @@ aggregation and display of all the components in the Application.
         <td>string</a></td>
         <td>A short, human readable textual description of the Application.</td>
     </tr>
+    <tr>
+        <td>spec.icon</td>
+        <td>string</a></td>
+        <td>A url to the location for an image (e.g., png, gif, or jpeg) representing the Application.</td>
```
konryd: Can we call out data urls?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

deustis: +1 on supporting Data URIs. External URLs need to be fetched by the end-user's browser. Fine for cases where the image asset can be hosted publicly, but adds complexity when the app is not intended to be public.

deustis:
@mattfarina, you were asking above about icon requirements in the Google Cloud console. I work on the [Cloud Launcher](http://console.cloud.google.com/launcher) product. Here is the guidance we currently share with our partners:
- JPG or PNG format
- White or transparent background
- Aspect ratio between 1:1 and 2:1 (to support wide logos)
- Minimum of 130px per dimension
- File size: 5MB or smaller

However, these are onboarding guidelines. We typically then generate multiple pixel density and aspect ratio versions of the icon for use in different areas of the console.

deustis:
Other references:
https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/app-icon/
https://developer.android.com/guide/practices/ui_guidelines/icon_design.html
https://docs.microsoft.com/en-us/azure/marketplace-publishing/marketplace-publishing-push-to-staging

README.md
prydonius: +1 for this too

prydonius:
@deustis I like the idea of stating some guidelines here about icons. As you mentioned, if the tooling needs to, it can always process the images to generate different variants.

+1 on creating an imageSpec struct so we can extend it if we see the need in the future.

mattfarina:
@smarterclayton @jzelinskie Do you know anyone who'd want to give input on image icons as they are passed around in metadata?

keithguerin:
Hi, As a designer on Google Cloud Console and contributor to GKE+apps I can offer some advice.

While SVGs are in theory preferable, there are some issues to watch out for. Many providers do not have svg logos on hand. Some svgs have complex layering and blend modes resulting which result in rendering issues. Many have surprisingly large file size. For those reasons, we have requested high res bitmaps that we size down, as David described above.

Progressive web app manifests provide an open-source schema for providing web app icons. It might be reliable and easier to adopt their schema than create our own. https://developers.google.com/web/fundamentals/web-app-manifest/

README.md
jzelinskie: How this is supported in Tectonic's system for operators is with the following JSON Schema:

```yaml
icon:
    type: array
    description: Icon which should be rendered with the application information
    required:
    - base64data
    - mediatype
    properties:
    base64data:
        type: string
        description: Base64 binary representation of the icon image
        pattern: ^(?:[A-Za-z0-9+/]{4}){0,16250}(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$
    mediatype:
        type: string
        description: Mediatype for the binary data specified in the base64data property
        enum:
        - image/gif
        - image/jpeg
        - image/png
        - image/svg+xml
```

We actually avoid remote hosted URLs for cross-site request reasons, if you were to display these in a dashboard.

deustis: @jzelinskie, I'm wondering, why not follow the data URI syntax where media type and encoded data are encoded in a single string?

The advantage of conforming to URI syntax is that dashboards can then choose to implement other mechanisms for fetching images: external URLs, storage buckets, or perhaps other intra-cluster storage mechanisms.

jzelinskie: We explicitly wanted to limit what could be acceptable input as an icon. This way a dashboard knows it has a finite number of things that it needs to support in order to support all icons. It was us choosing to be slightly more opinionated in order to make building tools that support the resource easier.

mattfarina:
This one is complicated. What the Kubernetes project puts out is trying not to be opinionated on workflow yet we want to enable tool interoperability. It's hard to walk the line of how far to go into opinions. I've added this to the sig apps meeting agenda to discuss.

mattfarina:
The update here is based on [Web App Manifests icons](https://www.w3.org/TR/appmanifest/#icons-member).

jzelinskie:
re: web app manifests:

>Implementors need to be aware that this specification is not stable. However, aspects of this specification are shipping in at least one browser (see links to implementation status at the top of this document). Implementors who are not taking part in the discussions will find the specification changing out from under them in incompatible ways. Vendors interested in implementing this specification before it eventually reaches the Candidate Recommendation phase should subscribe to the repository on GitHub and take part in the discussions.

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -78,6 +82,23 @@ type ApplicationStatus struct {
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 }
 
+// ImageSpec contains information about an image used as an icon.
+type ImageSpec struct {
+
```
jzelinskie: this newline doesn't exist on any other type. yay for nitpicking

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -78,6 +82,23 @@ type ApplicationStatus struct {
...
+	Source string `json:"src"`
+
+	// (optional) The height of the image in pixels.
+	Height string `json:"height,omitempty"`
+
+	// (optional) The width of the image in pixels.
+	Width string `json:"width,omitempty"`
+
+	// The mine type of the image (e.g., "image/png").
+	Type string `json:"type,omitempty"`
```
prydonius: is this optional?

jzelinskie: Data URLs are going to contain it, so it may be unnecessary in that case.

pkg/apis/app/v1alpha1/application_types.go
mattfarina: Removed extra line. Thanks for catching the consistency issue.

pkg/apis/app/v1alpha1/application_types.go
mattfarina: Added the option note to the comment.

mattfarina:
@kow3ns @prydonius I updated the PR per our chat today.

This includes changes height and width to size like the web app manifest

README.md
```diff
@@ -202,6 +207,7 @@ spec:
   descriptor:
     version: "4.9.4"
     description: "WordPress is open source software you can use to create a beautiful website, blog, or app."
+    icon: "https://example.com/wordpress.png"
```
ant31: should it be:
```
 icons:
   - src: "https://example.com/wordpress.png"
     type: "image/png"
```

docs/example.yaml
```diff
@@ -16,6 +16,9 @@ spec:
       kind: StatefulSet
   version: "4.9.4"
   description: "WordPress is open source software you can use to create a beautiful website, blog, or app."
+  icons:
```
ant31: I would include multiple icons and the size field in the example

mattfarina: @ant31 FYI, I plan on opening a second PR that introduces the purpose field. Doing it as a separate PR so we can discuss/debate that single point. purpose is part of the web app manifest... https://www.w3.org/TR/appmanifest/#dom-imageresource-purpose

mattfarina: Adding with real examples.

docs/example.yaml
```diff
@@ -16,6 +16,13 @@ spec:
...
+    - src: "https://s.w.org/style/images/about/WordPress-logotype-wmark.png"
+      type: "image/png"
+      size: "1000x1000"
+    - src: "https://s.w.org/style/images/about/WordPress-logotype-standard.png"
```
jzelinskie: Is this something we can legally use?

mattfarina: We are referencing a URL rather than using the image. If someone uses the example it is upon them to make sure they follow the policy when they use the images. Is there a problem with referencing these URLs? Something I'm not aware of.

mattfarina:
@kow3ns The latest update is based on master with v1beta1 and squashed into a single commit.

adambkaplan:
There's a concern that if `src` is a URI, it won't work well for folks running offline clusters. Folks who deploy applications without an internet connection will be faced with an uncomfortable choice:

1. Embed the icon in a data URI (security and usability concerns).
2. Mirror icons somewhere on-cluster.

We've been exploring using ConfigMaps to store assets on-cluster (SVG in particular), then enabling the contents to be exposed publicly (allowing tool-agnostic access)

mattfarina:
@adambkaplan Please note, this supports Data URLs can can contain the complete image.

Issue #11, opened 2018-03-26
Application status


konryd:
The Application resource is intended to be used by UIs. One of the most useful features of those is reporting the health of the displayed resources.

The "status" of the application (for the sake of discussion, let's simplify to either "OK" or "Error") can be generated in one of the following ways:
1. With an explicit field in ApplicationStatus.
1. With a reference to a a running pod's readiness probe.
1. By analyzing Kubernetes-level events involving the application.
1. By computing status of each of the Application's components and using the most problematic one.

Each of the approaches carries some trade-offs:

### Explicit field in ApplicationStatus
Pros:
1. Easy to consume.

Cons:
1. Would have to be populated by the application controller, the logic that would run there is far from obvious and might differ depending on application

### With a reference to a a running pod's readiness probe.
Pros:
1. Easy to consume.
1. Easy to implement custom status logic.
1. The pod to use might be one that already runs inside the app (eg. the worpress app might use the serving pod for this, which would also check the database)

Cons:
1. The pod would have to be managed with a deployment, growing the overhead for each application.
1. The application owners / providers would have to include the readiness pod.
1. Puts additional load on cluster resources.

NOTE: this could be an optional approach, with a default fallback being any of the other approaches.

### By analyzing Kubernetes-level events
Pros:
1. Implementable using only existing Kubernetes mechanism.
1. Similar approach works well for existing resources ([eg. Deployment in kube-dashboard](https://github.com/kubernetes/dashboard/blob/3988cad30bba696e1382159e6e00a64ec1bf298c/src/app/backend/resource/deployment/list.go#L137))

Cons:
1. The algorithm for changing a list of events into a status would have to wear out in use and will be easy to break, especially in the initial period until the Application resource matures.
1. Most likely, will only be able to capture the Kubernetes-control-plane-level problems. Making it work for individual component's statuses as well will inherit all of the other solutions' problems.
1. Will force each consumer to re-implement the status computation logic.

### By computing status of each of the Application's components and using the most problematic one
Pros:
1. Exhaustive and very likely to match users' expectations.

Cons:
1. Expensive to compute, even more so when listing all applications in cluster (where it would pretty much mean querying for all available GroupKinds everywhere).

konryd:
> events are not a good way to determine current state. They can be lost, and they can conflict, and they are not required to be delivered or emitted
Aren't these worries theoretical? The events are used to generate pod's status eg. in Kubernetes Dashboard. I am not not aware of users complaining.

[...]

PR #19, opened 2018-04-12
Added field "phase" to the ApplicationSpec

konryd:
The field is to be used by deployers (either automated or manual) to indicate whether all of application's resources have already been deployed.

### DIFF
--- a/pkg/apis/app/v1alpha1/application_types.go
+++ b/pkg/apis/app/v1alpha1/application_types.go
@@ -58,6 +58,9 @@ type ApplicationSpec struct {
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
 	Notes string `json:"notes,omitempty"`
+
+	// Phase represents the current phase of the application.
+	Phase ApplicationPhase `json:"phase,omitempty"`
 }
 
 // ApplicationStatus defines controllers the observed state of Application
@@ -97,6 +100,17 @@ type InfoItem struct {
 	// Value is human readable content.
 	Value string `json:"value,omitempty"`
 }
+
+type ApplicationPhase string
+
+const (
+	// Used to indicate that not all of application's components
+	// have been deployed yet.
+	Assembly ApplicationPhase = "Assembly"
+	// Used to indicate that all of application's components
+	// have alraedy been deployed.
+	Assembled ApplicationPhase = "Assembled"
+)
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/hack/install.yaml
+++ b/hack/install.yaml
@@ -79,6 +79,11 @@ spec:
               type: array
             description:
               type: string
+            phase:
+              type: string
+              enum:
+              - Assembly
+              - Assembled
             info:
               items:
                 properties:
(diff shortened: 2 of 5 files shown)


konryd:
I'm open to ideas for better names (yeah, I don't really like these either), but I don't think the names of Pod phases can apply to the semantic needs here.

In Issue #11 we're considering two families of solutions - health checks and component status aggregation. Both of those approaches, however, are not able to detect cases when application is still being constructed, that is: the Application resource is already there, but its components are not.

The "phase" field is intended to enable deployers/installers to provide that missing part of information (note that it resides in the ApplicationSpec and not status). Once there, it might be used to improve any specific solution for detecting status.

prydonius:
I don't think we need to stick to the exact semantics Pods use here, I was merely advocating for using common status names. We'd still need to document the exact meaning of each phase.

*Pending* - App resource is there, but components are not
*Succeeded* - App resource and components are there (note: this does not indicate that app is healthy)
*Failed* - App resource is there, but there was an error when creating a component (not entirely sure this makes sense, but I'm thinking about a case where a `helm install` failed because of a manifest error - e.g. a Container missing the image field)

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -98,6 +101,17 @@ type InfoItem struct {
...
 
+type ApplicationPhase string
+
+const (
+	// Used to indicate that not all of application's components
+	// have been deployed yet.
+	Assembly ApplicationPhase = "Assembly"
+	// Used to indicate that all of application's components
+	// have alraedy been deployed.
+	Assembled ApplicationPhase = "Assembled"
```
kow3ns: You don't need provide the type for every const

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -58,6 +58,9 @@ type ApplicationSpec struct {
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
 	Notes string `json:"notes,omitempty"`
+
+	// Phase represents the current phase of the application.
+	Phase ApplicationPhase `json:"phase,omitempty"`
```
kow3ns: Not sure we should use Phase in Spec. I think we may want to reuse this in Status. Also, no other types afaict use a Phase in the Spec.

We could put Phase in the Status, but then the client would be updating the Status, and this breaks with k8s convention. Also, based on what you are doing I agree that Spec is the right place for it.

If we are only ever going to have two phases though should we consider 
```go
ComponentsDeployed bool `json,"componentsDeployed,omitempty"`
```
or 
```go
Assembled bool `json,"assembled,omitempty"`
```
If we think there are going to be additional phases we could also, to disambiguate from typical Kubernetes Phases in Status objects, do something like
```go
AssemblyPhase ApplicationAssemblyPhase `json,"assemblyPhase,omitempty"`
```

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -98,6 +101,17 @@ type InfoItem struct {
 	Value string `json:"value,omitempty"`
 }
 
+type ApplicationPhase string
+
+const (
+	// Used to indicate that not all of application's components
+	// have been deployed yet.
+	Assembly ApplicationPhase = "Assembly"
```
kow3ns: I do like the idea of changing the first phase to pending.

hzxuzhonghu:
I can not find why put this **`AssemblyPhase`** in spec not Status? Could anyone explain it ?

jdmaguire:
I have to second @hzxuzhonghu, it's quite an anomaly to see this in spec and not in Status.

PR #20, opened 2018-04-17
Add back definition of CRD acronym

barry-schneider:
The acronym CRD is no longer defined in the document. The original version included defined the acronym the first time CRD (Custom Resource Definition) was used in the document, which is customary.


barry-schneider:
I have attached the [signed CLA](https://identity.linuxfoundation.org/system/files/clas/signed/cla_individual-28308ef7f87aa3d2c3ed4807a9b1263f50b22536.pdf) for your review

PR #34, opened 2018-04-25
Fixed the example application manifest

liyinan926:


barry-schneider:
Are these changes incorporated in the example under docs?

https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml

liyinan926:
Duplicated with #35.

PR #35, opened 2018-04-26
Fixed the example app with fixes from PR#4

konryd:
The componentKinds live under spec, not metadata


barry-schneider:
Are these changes incorporated in the example under docs?

https://github.com/kubernetes-sigs/application/blob/master/docs/example.yaml

PR #39, opened 2018-04-30
Update e2e-prow.sh to change target test repo

foxish:
cc/ @kow3ns


foxish:
We should also be able to test the automatic merge with an lgtm & approved label here.

Issue #12, opened 2018-03-26
Referencing Secrets and other resources


konryd:
This issue captures one of the threads on #4 where it was proposed to allow the Application to reference values/properties from Secret, ConfigMap, Ingress or Service.

Some quotes from the thread:
From @deustis: "[...]On the topic of credentials, we might want to establish a pattern for referencing Secrets. A very common getting started experience for an off-the-shelf app is trying to locate the password for an admin dashboard[...]"

@huyhg proposed simple templating system, where the template itself would live in spec and interpolated template would live in status.

@prydonius prefers enumerating reference types explicitly (Secret, ConfigMap, Ingress, Service)

@deustis proposed a templating system that would live in in spec but the evaluation would be the client responsibility (like @huyhg's proposal, but with no interpolation into spec). Also, the individual fields of the referenced objects would be selected using jsonpath.

@konryd pointed out that secret's values are typically evaluated on startup.

konryd:
@deustis - to follow on my point - The assumption that applications support reloading of Secrets after editing is a big one. There's more info here (search for config maps):

https://github.com/kubernetes/kubernetes/issues/33664

huyhg:
The discussions were specifically around information in `Value`s of the `Info` key-value pairs. The observation was that the values typically come from some common resources like Secret, ConfigMap, Ingress, or Service. We discussed various ways of how such references can be achieved and templatized.

deustis:
Thanks @konryd for capturing the context from the thread. I'll just add a couple of the concrete proposals that were mentioned there:

In the model proposed by @prydonius where reference types are enumerated, this would be an example for a Secret reference:

```
- name: username
  value: user
- name: password
  valueFrom:
    secretKeyRef:
       name: wordpress-secret
       key: wordpress-password
```

Alternative, using more generalized resource references, this might look like:

```
- name: username
  value: user
- name: password
  valueFrom:
    kind: Secret
    name: wordpress-secret
    jsonPath: $.data.wordpress-password
    base64Decode: true
```

We might also consider supporting templatized values, which could either be evaluated by the UI or by a controller associated with the Application resource.

For example, to build an admin URL with dynamic IP and port assignments:

```
- name: Admin URL
  templateValue: http://{{ serviceIp }}:{{ servicePort }}/admin
  templateData:
    serviceRoot:
      valueFrom:
        kind: Service
        name: wordpress-external-service
        jsonPath: $.spec.clusterIP
    servicePort:
      valueFrom:
[...]

PR #41, opened 2018-05-10
Support references in InfoItem

huyhg:
Implements proposal in #12.


hack/install.yaml
```diff
@@ -68,30 +68,45 @@ spec:
...
+                  valueFrom:
+                    properties:
+                      ingressRef:
+                        properties:
+                          host:
+                            type: string
+                          path:
+                            type: string
+                        type: object
+                      secretRef:
```
prydonius: I think this should be `secretKeyRef` to match what's used in the podspec. Also there should be a `configMapKeyRef` too, right?

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,42 @@ type InfoItem struct {
 
 	// Value is human readable content.
 	Value string `json:"value,omitempty"`
+
+	// Value is derived from another source.
+	ValueFrom *InfoItemSource `json:"valueFrom,omitempty"`
+}
+
+type InfoItemSource struct {
+  SecretKeyRef *ConfigMapKeySelector `json:"secretRef,omitempty"`
```
prydonius: could we not just reuse `core.SecretKeySelector`?

hack/install.yaml
huyhg: Good catch, thanks! I missed updating the json annotation.

pkg/apis/app/v1alpha1/application_types.go
huyhg: Thanks for the pointer. Yes, we can re-use it. Updated!

hack/install.yaml
```diff
@@ -68,30 +68,42 @@ spec:
...
+                      configMapKeyRef:
+                        type: object
+                      ingressRef:
+                        properties:
+                          host:
+                            type: string
+                          path:
+                            type: string
+                        type: object
+                      secretKeyRef:
```
huyhg: @kow3ns I'm using the latest kubebuilder, but it doesn't seem to be able to pull in the definitions of core structs here.

hack/install.yaml
```diff
@@ -186,9 +182,50 @@ metadata:
...
+        resources:
+          limits:
+            cpu: 100m
+            memory: 30Mi
+          requests:
+            cpu: 100m
+            memory: 20Mi
+      terminationGracePeriodSeconds: 10
+  updateStrategy: {}
+status:
```
huyhg: @kow3ns not sure what's going on with the generated code here...

huyhg:
@konryd FYI

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,34 @@ type InfoItem struct {
 
 	// Value is human readable content.
 	Value string `json:"value,omitempty"`
+
```
kow3ns: You are implementing the golang version of a union. You need to add a type specified to indicate whether Value or ValueFrom is set e.g. 
```go
type InfoItemType string

const (
  Value InfoItemType = "value"
  Reference = "reference"
)
```
You may also want to make Value a pointer. Only one should be set, and it should correspond to the indicated type field.

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,34 @@ type InfoItem struct {
...
+type InfoItemSource struct {
+	SecretKeyRef    *core.SecretKeySelector    `json:"secretKeyRef,omitempty"`
+	ConfigMapKeyRef *core.ConfigMapKeySelector `json:"configMapKeyRef,omitempty"`
+	ServiceRef      *ServiceSelector           `json:"serviceRef,omitempty"`
+	IngressRef      *IngressSelector           `json:"ingressRef,omitempty"`
+}
+
+type ServiceSelector struct {
+	// The Service to select from.
+	core.LocalObjectReference `json:",inline"`
```
kow3ns: If you use LocalObjectReference the Service must appear in the same namespace as the Application. The same is true of Ingress. Is this only meant to work with load balanced services?

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,34 @@ type InfoItem struct {
 
 	// Value is human readable content.
 	Value string `json:"value,omitempty"`
+
+	// Value is derived from another source.
+	ValueFrom *InfoItemSource `json:"valueFrom,omitempty"`
+}
+
+type InfoItemSource struct {
```
kow3ns: InfoItemSource, ServiceSelector, and IngressSelector need comments.

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,34 @@ type InfoItem struct {
...
+	SecretKeyRef    *core.SecretKeySelector    `json:"secretKeyRef,omitempty"`
```
kow3ns: You want need to have a type field to indicate the member of the union that contains the value. The same as InfoItem above. Each of the values needs a comment to indicate the use.

pkg/apis/app/v1alpha1/application_types.go
huyhg: Understood your comment about union in Go. Such type wouldn't be consistent with what is currently in the core API though. I'm drawing from the following examples:

https://github.com/kubernetes/api/blob/184e700b32b7f1b532b9fce8dd8c1f412d297c4b/core/v1/types.go#L1663

pkg/apis/app/v1alpha1/application_types.go
huyhg: Similar to above. I'm drawing from the following example in core:
https://github.com/kubernetes/api/blob/184e700b32b7f1b532b9fce8dd8c1f412d297c4b/core/v1/types.go#L1681

pkg/apis/app/v1alpha1/application_types.go
huyhg: Done.

pkg/apis/app/v1alpha1/application_types.go
huyhg: Changed to `core.ObjectReference`. Discovered that `core.ConfigMapKeySelector` and such also use `LocalObjectReference`, so I created new `ConfigMapKeySelector` and `SecretKeySelector`.

huyhg:
PTAL.

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -100,6 +101,57 @@ type InfoItem struct {
...
+type ConfigMapKeySelector struct {
+	// The ConfigMap to select from.
+	core.ObjectReference
+	// The key to select.
+	Key string
+}
+
+// SecretKeySelector selects a key from a Secret.
+type SecretKeySelector struct {
+	// The Secrete to select from.
```
prydonius: Secret

foxish:
@kow3ns, can you retry the `/approve`?

huyhg:
Can we merge this change now please?

foxish:
@kow3ns needs /lgtm as well.

huyhg:
@konryd FYI, this has been merged.

PR #45, opened 2018-06-06
Add Descriptor field  

### LINKED ISSUE #28: Nest Application metadata fields
I'd separate the metadata fields about an application, like `description, keywords...` from the fields holding some logic like `matchLabels`, `componentKinds`. 

First, it's more clear for the user and tools authors to spot what is informative only and what is modifying the behavior of the controller 
Second, the descriptor spec could be more easily reused by other project and integrated to different CRD.

related comment (from the kinflate initial PR): https://github.com/kubernetes/kubernetes/pull/52570#discussion_r140929592
 
``` go
  type Application struct {
     
     // Metadata about the app. 
     descriptor Descriptor            
     ...
 }

  // Descriptor is a convenience struct gathering metadata to support
  // Manifest search, browse and updates.
  type Descriptor struct {
     Name string
     Version string
     Description string
     Icon string
     Keywords []string
     Homepage string
     ...
  }
```

ant31:
fixes #28 

few questions: 
1. How hack/* is generated? (the CRD spec and examples) ?
1. How the readme fields-table is generated  ?
1. Is the `version` field  going to be used by the controller (e.g to rollout apps) ?


README.md
```diff
@@ -199,25 +198,27 @@ spec:
...
-    - "blog"
-    - "wordpress"
-  links:
-    - description: About
-      url: "https://wordpress.org/"
-    - description: Web Server Dashboard
-      url: "https://metrics/internal/wordpress-01/web-app"
-    - description: Mysql Dashboard
-      url: "https://metrics/internal/wordpress-01/mysql"
+  applicationMetadata:
```
deustis: "application" in the field name seems redundant since. Would also suggest something like "descriptor" rather "metadata" to highlight the purpose of this object (descriptive information). Rename field to "descriptor"?

prydonius: +1 for "descriptor"

ant31: Yes, initially it was 'descriptor'. I change it back to it

konryd:
Apologies for the delay. I think this change is desireable and this is the best time to make it :) Thanks!

ant31:
I've updated the sample and documentation. 
Would be nice to have them generated, at least the `hack/install.yaml` in the futur

PR #48, opened 2018-06-15
Soft-suggesting markdown for formatting

konryd:
The textual fields will inevitably need some formatting (lists, code samples etc.).

Rather than forcing any specific markup language, the Application type should suggest using one that's possible to render
in plain text, with a nod towards markdown as example of this.

### DIFF
--- a/pkg/apis/app/v1alpha1/application_types.go
+++ b/pkg/apis/app/v1alpha1/application_types.go
@@ -38,6 +38,7 @@ type ApplicationSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// Description is a brief string description of the Application.
+	// Should contain plain text or plain-text-friendly markup (eg. markdown).
 	Description string `json:"description,omitempty"`
 
 	// Maintainers is an optional list of maintainers of the application. The maintainers in this list maintain the
@@ -58,6 +59,7 @@ type ApplicationSpec struct {
 	Links []Link `json:"links,omitempty"`
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
+	// Should contain plain text or plain-text-friendly markup (eg. markdown).
 	Notes string `json:"notes,omitempty"`
 
 	// AssemblyPhase represents the current phase of the application's assembly.


prydonius:
What about soft-requiring it to be Markdown? The issue I see with suggesting Markdown as an example of plain-text friendly markup, is that someone could instead choose to use reStructuredText, and now a UI has the job of figuring out which one is being used. Plain-text is valid Markdown, so just describing that the field should be Markdown is enough IMO (but describing the field should be "either plain-text or Markdown" is also OK).

I would argue that Description should specifically not be formatted text as it is supposed to be brief. For comparison, in charts we only expect the README to be Markdown, description and notes are expected to only be plain-text. I can see the value in Notes being formatted, however.

konryd:
I agree with this suggestion. I'd rather the tools didn't have to sniff for the markup language. I'm glad to see the idea to bless a markup is welcome.

prydonius:
@konryd did you want to update this PR to address the above?

jzelinskie:
The OLM manifests supported Markdown in the description for a while and we used them to render a large portion of the page on the Console. It was nice, but not all tools in the pipeline are going to support rendering markdown. Importantly, if you want to front your applications with an Open Service Broker, that spec isn't going to be able to leverage your description if it's formatted in markdown.

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -38,6 +38,7 @@ type ApplicationSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// Description is a brief string description of the Application.
+	// Should contain plain text or plain-text-friendly markup (eg. markdown).
```
mattfarina: The idea of this is a simple string description rather than something long form. Here are examples from other platforms:

node.js (forever app):
> A simple CLI tool for ensuring that a given node script runs continuously (i.e. forever)

composer/packagist (guzzle library):
> Guzzle is a PHP HTTP client library

helm (artifactory chart):
> Universal Repository Manager supporting all major packaging formats, build tools and CI servers.

If we look at something like Ubuntu snaps there's a one like summary and a multi-line text description. One is meant for a sentence and the other a paragraph.

This is the idea for description. Why would it being from being in Markdown?

pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -58,6 +59,7 @@ type ApplicationSpec struct {
 	Links []Link `json:"links,omitempty"`
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
+	// Should contain plain text or plain-text-friendly markup (eg. markdown).
```
mattfarina: Note can be longer form. The things that could consume this information include CLIs, desktop apps, web apps, and more. By hinting at markup like markdown without saying it we could end up with the case of markdown, asciidoc, and others being used. Then tools from varying places need to know how to handle this for interoperability.

How do you see that playing out?

pkg/apis/app/v1alpha1/application_types.go
prydonius: Agreed, as suggested above, I think this should be plain-text only

pkg/apis/app/v1alpha1/application_types.go
prydonius: @mattfarina see my comment https://github.com/kubernetes-sigs/application/pull/48#issuecomment-397711962, it sounded like @konryd agreed with this and was thinking of changing this to require either plain-text or Markdown (or just Markdown since plain-text is valid Markdown).

mattfarina: @prydonius I was more thinking of what @jzelinskie [noted](https://github.com/kubernetes-sigs/application/pull/48#issuecomment-398871192):

> not all tools in the pipeline are going to support rendering markdown. Importantly, if you want to front your applications with an Open Service Broker, that spec isn't going to be able to leverage your description if it's formatted in markdown.

Supporting markdown means we are adding a dependency on all tools that display this to support markdown. How does adding that dependency play out in practice?

prydonius: I was referring to this comment:
> we could end up with the case of markdown, asciidoc, and others being used. Then tools from varying places need to know how to handle this for interoperability.

As for CLI and other tooling that may not want to render Markdown, I agree that might be a good reason to keep this plain-text also. The reason we explicitly chose for NOTES.txt in Helm to be plain-text is because the main use-case was a terminal emulator where Markdown formatting is not used (that said there are tools to improve rendering in terminal emulators: https://github.com/samfoo/mdcat). That said, I think Markdown generally is readable even when not rendered, and it may be helpful for authors to use it for Notes.

konryd: Apologies for the delay. Like @prydonius says - I'm about to change this to blessing markdown more explicitly.

I don't agree that this introduces a dependency for the tools - the idea behind markdown is to stay readable in the undrendered form. Notes rendered plainly should stay reasonably readable for humans. Out of all markup lanugages, this property of markdown is the reason to bless it. Otherwise, we'll be deciding against any sort of markup in the first place, which won't work well with UIs.

pkg/apis/app/v1alpha1/application_types.go
konryd: SG. I'm reverting this change.

konryd:
Update: this branch was created from github's website, which resulted in a fork that lives in the repo. Moved to my own fork now, but it requires a separate PR. It is here: https://github.com/kubernetes-sigs/application/pull/53

PR #52, opened 2018-07-02
V1beta1

kow3ns:
updates to the latest kubebuilder version and updates the resource to v1beta1.
The first commit only updates the dependencies in vendor.
The second commit should get human eyes for reveiw.


dlorenc:
Took a look over the second commit, changes look good to me.

/lgtm

docs/example.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: app.k8s.io/v1alpha1
+apiVersion: app.k8s.io/v1beta1
```
prydonius: should the `app.k8s.io` prefix be there? It was removed here https://github.com/kubernetes-sigs/application/pull/52/commits/18cbb53830259d38a1f7bbe77142ba7810b9f09a#diff-04c6e90faac2675aa89e2176d2eec7d8R183

README.md
```diff
@@ -180,7 +180,7 @@ The Application object shown below declares that the Application is a WordPress
 and Services. It also contains some other relevant metadata described above.
 
 ```yaml
-apiVersion: app.k8s.io/v1alpha1
+apiVersion: v1beta1
```
prydonius: the `app.k8s.io` prefix needs to be added back

jzelinskie:
Am I missing something do we really want to change this to beta before there's anything using it yet? We can't make any breaking changes after that point.

konryd:
@jzelinskie Application CRD does not package any controller and its value is in providing a stable interface. The exact shape has been iterated on for the last six months, and there are project underway that make use of it (eg. https://github.com/kubernetes/dashboard/pull/2980).

jzelinskie:
@konryd I never mentioned anything about a controller. This project is clearly still being iterated on as evident by the open PR for icon schema. Adding support to show the CRD in the dashboard is not evidence of any actual usage of this CRD.

deustis:
This won't be any surprise to @kow3ns (as we work together), but I'd like to also voice my support for moving the CRD to Beta --

The spec has been taking shape over the past six months. During that time, breaking changes such as #45 (by @ant31), have been accepted. Some stability is needed now to achieve the stated goal interoperability among tools and UIs. Extensions such as the icon PR (#17) can still be accommodated after the CRD goes to Beta.

konryd:
@jzelinskie The reason I mentioned the controller is to contrast the "CRD&controller working together and doing things" pattern with the purely declarative way in which the App CRD is used. 

As for the iteration - the Kubernetes API definition of Beta [1] does not imply a lockdown on the changes. In fact, it allows even changes that break backwards compatibility, which additive changes don't do.

[1] https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning

kow3ns:
@mattfarina @prydonius @ant31 can I get a LGTM?

PR #53, opened 2018-07-05
Encouraging using markdown for notes

konryd:
This is an updated version of https://github.com/kubernetes-sigs/application/pull/48 that lives in my own fork rather than a fork in the upstream repo.

Addressed the comments:

1. Dropped suggestion from the decription field.
1. Made the encouragement for explicit.

@mattfarina (github doesn't let me add you to reviewers)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ aggregation and display of all the components in the Application.
     <tr>
         <td>spec.descriptor.notes</td>
         <td>string</td>
-        <td>Notes contain a human readable snippets intended as a quick start for the users of the
-        Application.</td>
+        <td>Notes contain human readable snippets intended as a quick start for the users of the
+        Application. They may be plain text or <a href="spec.commonmark.org">CommonMark</a> markdown.</td>
     </tr>
     <tr>
         <td>spec.assemblyPhase</td>
--- a/pkg/apis/app/v1beta1/application_types.go
+++ b/pkg/apis/app/v1beta1/application_types.go
@@ -51,6 +51,7 @@ type Descriptor struct {
 	Links []Link `json:"links,omitempty"`
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
+	// CommonMark markdown syntax may be used for rich text representation.
 	Notes string `json:"notes,omitempty"`
 }
 


pkg/apis/app/v1alpha1/application_types.go
```diff
@@ -58,6 +58,7 @@ type ApplicationSpec struct {
 	Links []Link `json:"links,omitempty"`
 
 	// Notes contain a human readable snippets intended as a quick start for the users of the Application.
+	// Should contain plain text or markdown-formatted text.
```
mattfarina: Can you note using [CommonMark](http://commonmark.org/) as the specific kind and include the version (latest). If you search in the [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/fb77181150d665cef2257a1f9a5f398aa2404031/versions/3.0.0.md) you can see examples of how they phrased it.

konryd: Thanks for the link, this is very useful! I'm going to grab some of the wording but 'll drop the RFC-style capitalization.

bacongobbler:
Personally I'm 👎 on the current verbiage, as it implies that kubectl and other api-facing clients (Helm included) are forced to support only markdown and plain text, and not reST or other markdown languages (BBCode, for example).

I also kinda prefer the lazy approach for now: If we say that only plain text is supported, then that makes the implementation much simpler for clients (kubectl, Helm, etc).

Regardless of my opinion, if we plan on opening the spec to allow markdown syntax we should either choose to force api-facing clients to support all markdown languages, or none of them. I don't think that kubectl has support for rendering plain text as markdown/reST/BBCode at this time, so that would mean new work for SIG CLI. Is anyone willing to take up that work?

prydonius:
>  forced to support only markdown and plain text, and not reST or other markdown languages (BBCode, for example).

@bacongobbler one of the things we should avoid is having this too open-ended and requiring tooling to figure out if something is markdown, reST, BBCode or something else. That, to me, would be worse than forcing all tools to support only plain text/markdown.

mattfarina:
@bacongobbler If we leave it open ended there is no spec for what tool developers need to support. That's why the suggestion for CommonMark (a spec) instead of Markdown (an implementation that lead to flavors). I would suggest anything we do support ends up being backed by a spec.

There were some questions on github flavor markdown. They now use commonmark. Here is the blog post with the details... https://githubengineering.com/a-formal-spec-for-github-markdown/ 

I would suggest we either support specific spec'd things or plain text.

ant31:
I'm in favor to be opinionated here, plain-text or common mark seems a good way to go. 
Also it's a suggestion, if a company/oss wants to use reST or anything they can
if it's open-ended, we should add a field to indicate the fomat:

```
// Structured syntax may be used for rich text representation.
Notes string `json:"notes,omitempty"`

// Notes format
NotesFormat string `json:"notesFormat,omitempty"`
```

deustis:
Mostly chiming in to support @konryd's statement above. Starting with one opinionated format seems preferable for interoperability here, until we find strong reasons to branch out, and CommonMark seems a logical choice. 

I also like @ant31's proposal, probably not for this PR, but as a future extension mechanism should we feel the need support for additional rich text formats in the future.

keithguerin:
Agreed with konryd & deustis's comments: Opinionated format for interoperability (markdown is obvious choice due to wide adoption), with an eye towards future extensibility if needed.

mattfarina:
@ant31 Leaving it open ended means that all tools dealing with display need the ability to handle any possibility in there. CommonMark is already adding a dependency to tools.

ant31:
@mattfarina if it was not clear, I'm for limiting it to CommonMark. Such opinions, limited choices help more the community than the opposite. 
I'd re-open this discussion only if there is a push from the community for other formats.

konryd:
Can I get an LGTM?

konryd:
I also need "approve"

README.md
```diff
@@ -112,8 +112,8 @@ aggregation and display of all the components in the Application.
     <tr>
         <td>spec.notes</td>
```
ant31: should itnbe spec.descriptor.notes

PR #55, opened 2018-07-11
Replaced dependency on k8s.io/kubernetes/pkg/apis/core with k8s.io/api/core/v1

liyinan926:
`pkg/apis/app/v1beta1/application_types.go` had a dependency on `k8s.io/kubernetes/pkg/apis/core`, which caused the Kubernetes code base to be pulled into vendor. Replaced it with a dependency on `k8s.io/api/core/v1` instead.

@kow3ns


liyinan926:
@kow3ns Can you take a look?

PR #58, opened 2018-07-23
Offline Icons Schema

adambkaplan:
Allow icons to specify raw source data either as:
- binary base64 encoded string
- plain text (for SVG)
- reference to a ConfigMap entry

Work inspired in part by the [operator-lifecycle-manager](https://github.com/operator-framework/operator-lifecycle-manager), which includes a catalog that can be accessed on-cluster for air-gapped deployments.


pkg/apis/app/v1beta1/application_types.go
```diff
@@ -84,15 +84,36 @@ type ApplicationStatus struct {
...
-	// The source for image represented as either an absolute URL to the image or a Data URL containing
+	// Binary contains the base64-encoded image binary.
+	//
+	// Only one of Binary, Text, URL, or ConfigMapKeyRef may be defined.
+	Binary string `json:"binary,omitempty"`
+
+	// Text contains the image definition as plain text, for formats such as SVG.
+	//
+	// Only one of Binary, Text, URL, or ConfigMapKeyRef may be defined.
+	Text string `json:"text,omitempty"`
```
mattfarina: The current spec supports Data URLs ([IETF RFC 2397](https://tools.ietf.org/html/rfc2397)). These support base64 encoded images and SVG as text. CSS-Tricks has an article titled, ["SVG with Data URIs"](https://css-tricks.com/lodge/svg/09-svg-data-uris/) showing how these work.

Data URLs are, also, something that can be passed to browsers as they know how to handle them.

The Go URL knows how to parse Dara URLs. See example at https://play.golang.org/p/fAyOHuMNaxW

How will this change improve upon the Data URLs already in the codebase?

pkg/apis/app/v1beta1/application_types.go
```diff
@@ -84,15 +84,36 @@ type ApplicationStatus struct {
...
 	Size string `json:"size,omitempty"`
 
-	// (optional) The mine type of the image (e.g., "image/png").
+	// Type indicates the MIME type of the image (e.g., "image/png"). Optional.
 	Type string `json:"type,omitempty"`
+
+	// ConfigMapKeyRef is a reference to a ConfigMap entry containing the image data.
+	//
+	// Only one of Binary, Text, URL, or ConfigMapKeyRef may be defined.
+	ConfigMapKeyRef *ConfigMapKeySelector `json:"configMapKeyRef,omitempty"`
```
mattfarina: If someone were going to build a UI to display images in a configmap, how would they do it? A URL or Data URL are something that can be passed to a browser or other viewer. How would someone use a file in a configmap?

pkg/apis/app/v1beta1/application_types.go
prydonius: +1 Data URIs can be directly passed as a src to image tags, so I feel the existing field supports this

pkg/apis/app/v1beta1/application_types.go
adambkaplan: They would likely need to manually convert the info to a data URL - this is what the OpenShift (formerly Tectonic) console does under the covers. See [here](https://github.com/openshift/console/blob/master/frontend/public/components/cloud-services/index.tsx#L190).

konryd: Can we be more specific about the format of the ConfigMap value? Eg. "[...] containing JSON-formatted ImageSpec to use"

ant31: I don't see the advantages to do this via configmap, the definition of an application is not tied to be shipped with a configmap holding the image. The image could already be inserted as base64 data for offline usage.

adambkaplan:
Closing this PR, as the consensus is that data URI support is sufficient.
I also see the ConfigMap idea presenting more challenges than benefits/solutions.

PR #75, opened 2018-10-31
Update docs & examples to follow K8s recommended application labels

jessesuen:
Per K8s documentation, `app.kubernetes.io/instance` would be more appropriate to use as a selector for an application object to match its resources, as opposed to `app.kubernetes.io/name`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,9 @@ aggregation and display of all the components in the Application.
     <tr>
         <td>spec.selector</td>
         <td><a href=https://kubernetes.io/docs/concepts/overview/working-with-objects/labels>LabelSelector</a></td>
-        <td>The selector is used to match resources that belong to the Application. All of the applications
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -190,12 +189,13 @@ kind: Application
...
   labels:
-    app.kubernetes.io/name: "wordpress-01"
+    app.kubernetes.io/name: "wordpress"
+    app.kubernetes.io/instance: "wordpress-01"
     app.kubernetes.io/version: "3"
 spec:
   selector:
     matchLabels:
-     app.kubernetes.io/name: "wordpress-01"
+     app.kubernetes.io/instance: "wordpress-01"
```
ant31: The uniqueness could be name+instance. there's not recommendation around that point 
e.g
```
app.kubernetes.io/name: "wordpress"
app.kubernetes.io/instance: "01"
```

I would match labels both.

Issue #76, opened 2018-10-31
Status Aggregation for Application CRD


barney-s:
## Background
Application CRD has a list of GroupKinds and a LabelSelector to logically group live objects in the cluster. The Application CRD also provides a way to attach descriptive metadata to the logical grouping. 

## What does this solve
This proposes to enhance Application CRD controller to provide status tracking for the objects that match the selectors. This dramatically simplifies the status tracking requirements for a client that uses Application CRD. 

## Proposal
The proposal is to add a reconciler that creates watches for the GroupKinds that the application CRD refers to. The reconciler then inspects the matching objects and updates the aggregate status in the Application CRD.

Please review the attached proposal: https://github.com/kubernetes-sigs/application/pull/77

barney-s:
@kow3ns please review the proposal.

enisoc:
@barney-s Could you copy the markdown into an .md file and make a PR to add that file to a new `docs/proposals/` dir in the repo? That would make it easier to make comments on specific parts.

barney-s:
> @barney-s Could you copy the markdown into an .md file and make a PR to add that file to a new `docs/proposals/` dir in the repo? That would make it easier to make comments on specific parts.

Done: https://github.com/kubernetes-sigs/application/pull/77

kow3ns:
I borrowed from the prior and came up with a new issue https://github.com/kubernetes-sigs/application/issues/78

barney-s:
#81 merged for this

PR #79, opened 2018-11-05
proposal for status

barney-s:
capturing proposal for review #76

### DIFF
--- a/docs/proposals/status.md
+++ b/docs/proposals/status.md
@@ -1,0 +1,126 @@
+## Background
+Application CRD has a list of GroupKinds and a LabelSelector to logically group live objects in the cluster. The Application CRD also provides a way to attach descriptive metadata to the logical grouping. 
+
+## What does this solve
+This PR proposes to enhance Application CRD controller to provide status tracking for the objects that match the selectors. This dramatically simplifies the status tracking requirements for a client that uses Application CRD. 
+
+## Proposal
+The proposal is to add a reconciler that creates watches for the GroupKinds that the application CRD refers to. The reconciler then inspects the matching objects and updates the aggregate status in the Application CRD.
+
+### New fields in Application CRD
+We propose to use Conditions to aggregate status of the matching objects.
+These conditions could be added to `.status.conditions`
+```
+"Ready"   //Aggregate readiness
+"Settled" //Observed generation is latest, the controller has acted on latest version of spec
+```
+In addition we could have the matching objects status also recorded in `.status.components`.  The motivation behind this is to provide individual component status as well for clients that need more than just the aggregate status. Well known object types can be aggregated to a higher degree of fidelity. For other types and custom resources we can propose standardizing `Ready` condition that could then be aggregated.
+
+```
+package application
+// ApplicationStatus defines controllers the observed state of Application
+type ApplicationStatus struct {
+	// ObservedGeneration is used by the Application Controller to report the last Generation of an Application
+	// that it has observed.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// Conditions - Ensure Ready,Settled conditions are set
+	Conditions []status.Condition   `json:"conditions,omitempty"`
+	// List of subcomponent status
+	Resources status.ResourceStatus `json:",inline"`
+}
+
+--------------------
+
+package status
+// ResourceStatus is a generic status holder for the top level resource
+type ResourceStatus struct {
+	// Object status array for all matching objects
+	Objects []ObjectStatus `json:"objects,omitempty"`
+}
+
+// ObjectStatus is a generic status holder for objects
+type ObjectStatus struct {
+	// Link to object
+	Link string `json:"link,omitempty"`
+	// Name of object
+	Name string `json:"name,omitempty"`
+	// Kind of object 
+	Kind string `json:"kind,omitempty"`
+	// Object group
+	Group string `json:"kind,omitempty"`
+	// Status. Values: InProgress, Ready, Unknown
+	Status string `json:"status,omitempty"`
+	// progress is a fuzzy indicator. Interpret as a percentage (0-100)
+	// eg: for statefulsets, progress = 100*readyreplicas/replicas
+	Progress int32 `json:"progress"`
+}
+
+type Condition struct {
+...
+}
+
+``` 
+
+### Logic for computing object status of well known objects
+The logic to compute the readiness for the objects is based on the resource group-kind.
+1. Statefulset is Ready if `.status.readyreplicas == .spec.replicas && .status.currentreplicas == .spec.replicas`
+2. Service is Ready always (could not derive from `.status.loadbalancer`)
+3. PodDisruptionBudget is Ready if .`status.currenthealthy >= .status.desiredhealthy`
+
+The long term plan is for these logic to be added to core-controllers and them to update the `.status.conditions.[Ready]` for the individual objects.
+
+### Status for other objects
+We propose looking for `.status.conditions.[Ready]` for non well known group-kinds and using that for aggregation.
+If the object does not have `.status.conditions.[Ready]`, we propose to mark that objects status as "Unknown". We will revisit this after feedback from users.
+
+### Aggregating status
+Application CRD's `.status.conditions.[Ready]` is derived by a simple ANDing of the matching object status computed under `.status.components`. 
+
+### Example output
+```
+status:
+  objects:
+    - progress: 100
+      link: /apis/apps/v1/namespaces/default/statefulsets/esbasic-di
+      name: esbasic-di
+      group: apps/v1
+      kind: statefulset
+      status: Ready
+    - progress: 100
+      link: /apis/apps/v1/namespaces/default/statefulsets/esbasic-m
+      name: esbasic-m
+      group: apps/v1
+      kind: statefulset
+      status: Ready
+    - progress: 100
+      link: /apis/apps/v1/namespaces/default/statefulsets/esbasic-metrics
+      name: esbasic-metrics
+      group: apps/v1
+      kind: statefulset
+      status: Ready
+    - progress: 100
+      link: /api/v1/namespaces/default/services/esbasic-di
+      name: esbasic-di
+      group: v1
+      kind: service
+      status: Ready
+    - progress: 100
+      link: /api/v1/namespaces/default/services/esbasic-m
+      name: esbasic-m
+      group: v1
+      kind: service
+      status: Ready
+  conditions:
+  - lastTransitionTime: 2018-10-29T23:30:03Z
+    lastUpdateTime: 2018-10-29T23:30:03Z
+    message: All components ready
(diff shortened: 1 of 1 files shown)


barney-s:
recreating a pull with the review comments of #77 addressed.
Since i created the origin pull directly on the browser, i lost the branch and the ability to update.
https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
So i ended up recreating a new pull request.

barney-s:
@dlorenc 
@mortent 
@deustis
@krmayankk
@huyhg 

Please review this proposal

docs/proposals/status.md
```diff
@@ -1,126 +1,300 @@
...
+
+#### Standard Conditions
+These conditions should be added to `.status.conditions` for Application CRD and subsequently to other resources:
+##### `Ready`
+Readiness as perceived by the resource controller. This means the controller deems the underlying resources to be ready for consumption to the best of its knowledge. It coule be implemented as a low pass filter to filter out transient or persistent failures such as restarts of underlying pods.
```
selfmanagingresource: s/coule/could/

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,1,opt,name=observedGeneration"`
+	// Conditions represents the latest state of the object
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,10,rep,name=conditions"`
+	// Resources embeds a list of object statuses
+	// +optional
+	ComponentList `json:",inline,omitempty"`
```
huyhg: `json:"components,inline,omitempty"`

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+	status.Meta `json:",inline"`	
+}
+```
+
+
+### Example
+```
+status:
+  components:
+  - link: /apis/apps/v1/namespaces/default/statefulsets/basic-di
```
huyhg: this example is not consistent with what described above. Each component should have a list of conditions instead of a single `status` field.

```
status:
  components:
  - link: /apis/apps/v1/namespaces/default/statefulsets/basic-di
    conditions:
    - type: Ready
      status: "True"
    - type: Settled
      status: "True"
```

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+#### Standard Conditions
+These conditions should be added to `.status.conditions` for Application CRD and subsequently to other resources:
+##### `Ready`
+Readiness as perceived by the resource controller. This means the controller deems the underlying resources to be ready for consumption to the best of its knowledge. It coule be implemented as a low pass filter to filter out transient or persistent failures such as restarts of underlying pods.
+
+##### `Settled`
+Would indicate whether controller acted on latest revision of `.spec` and the underlying resources conform to the latest revision of the `.spec`
+
+##### `Error`
+Would capture Errors seen in last reconcilation loop.
```
hausdorff: Which errors would be reported here? The proposal seems to suggest it would be "shallow" errors (_e.g._, "this particular `Deployment` failed to progress"), rather than the transitive closure of all errors in all resources managed by the app. CRD (_e.g._, "here are all the `Pod` crash errors that caused this particular `Deployment` to fail")?

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Fields
+Fields are a structured way to capture the status of resources. As part of this proposal, no standardized fields are proposed. The authors opinion is it does not provide additional information than what conditions provide. If the community strongly feels about standardizing some fields for in-tree and custom resources, it can be considered. 
+
+### Conditions
+[Conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#typical-status-properties) are used to describe the status of resources in an extensible way. Out of tree controllers support a variety of conditions. There is no consistent usage across the core controller. This is exacerbated with custom controllers. To tackle this diversity and its impact in status computation for Application CRD, we need a short term and a long term solution. Short term solution is to do aggregation of diverse core resources in the Application CRD controller. Long term is to identify best practices and standardize conditions and propagate them to the core controllers thereby making Application CRD aggregation simpler.
+
+#### Standard Conditions
```
hausdorff: Would be nice to have a formalized condition indicating whether this resource is awaiting GC, as well. Some resources default to bg deletion, and so on.

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+
+#### Other Conditions
+##### `Disruption`
+`Disruption` is a special case of `Error` where we may know what caused the error when `Ready` is false. It is useful for some resources and could be standardized as an optional condition. 
+
+### Aggregating status
+#### Custom Resources
+For custom resources, look for `Ready`, `Error` and `Settled`. If the object does not have the necessary conditions mark that objects status as "Unknown" and would not be used for computing the aggregate status. This would cover the case for nested Applciation CRDs as well. Based on feedback from users and reviewers this stance could be updated.
+
+### Core controller resources
```
hausdorff: For some resources, it is impossible to decide readiness from the resource definition itself. `Pod` is the best example of this, because (_e.g._) we don't know statically whether it is supposed to complete or not. It would be very nice to be able to display intent using annotations, or a similar mechanism. In the case of `Pod` itself, consider that Helm's test hooks allow us to launch `Pod`s that test whether an application is running -- so it is useful to (_e.g._) a monitoring/CI/CD system to communicate this intent, as it would allow it to wait for the tests to complete before progressing.

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+##### `Error`
+Would capture Errors seen in last reconcilation loop.
+
+#### Other Conditions from core controllers
+These are not as universal for all kubernetes objects as `Ready`, `Settled`, `Error` and provide yet another dimension of `Ready` and `Settled` condition. 
+
+##### `Progressing`
+Deployment uses this to indicate there is an update to the underlying ReplicaSet that is being acted upon. This is a inversion of the `Settled` condition.
+
+##### `Available`
```
hausdorff: I find `"Available"` is perpetually confusing name for this condition. People trying to build robust monitoring pipelines are nearly always caught off guard by this. It would be nice to report this as closer to what it really is -- succeeding liveness/readiness checks.

docs/proposals/status.md
barney-s: Thanks.
Do you prefer a set of conditions or a singular value ?

huyhg: I prefer set of conditions. My understandings of the proposal are:
- The proposed Ready and Settled types should eventually be universal on all resources.
- They will aggregate up to the Application's Ready and Settled types.
- It will take a while for core controllers to adopt these types.

So the application controller will enable users to derive Ready and Settled conditions of the standard resources now, which is very useful.

docs/proposals/status.md
barney-s: Would 'Ready' condition with '.status = false'  and set to '.reason = Finalizing' be sufficient ?
We need to also consider consistency. Deletion timestamp is set by API server synchronously. It is different from the controller setting the condition (potentially delayed). System relying on this condition may miss it during the delay.

docs/proposals/status.md
barney-s: ack

docs/proposals/status.md
barney-s: It depends on the controller design. We can make recommendations on how the error needs to be propagated up the hierarchy. But iam sure some controllers may end up hiding the errors. 
Its a good point and we can explicitly keep it as a design goal when propagating this condition to all core controllers.

docs/proposals/status.md
hausdorff: Are the values of  `.reason` intended to be common across all controllers? If so, I think that's fine. If not, I do think it merits its own condition, as it is a fundamental notion in Kubernetes, and it will only get _more_ fundamental as time goes on. (_c.f._, proposals that rollouts of `ConfigMap` are implemented using GC, _etc._)

docs/proposals/status.md
hausdorff: Yeah, I personally would much rather use annotations instead of shared `.status` mutation. :)

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+`Disruption` is a special case of `Error` where we may know what caused the error when `Ready` is false. It is useful for some resources and could be standardized as an optional condition. 
+
+### Aggregating status
+#### Custom Resources
+For custom resources, look for `Ready`, `Error` and `Settled`. If the object does not have the necessary conditions mark that objects status as "Unknown" and would not be used for computing the aggregate status. This would cover the case for nested Applciation CRDs as well. Based on feedback from users and reviewers this stance could be updated.
+
+### Core controller resources
+The logic to compute the readiness for the objects is based on the resource group-kind.
+1. Statefulset is Ready if `.status.readyreplicas == .spec.replicas && .status.currentreplicas == .spec.replicas`
+2. Service is Ready always (could not derive from `.status.loadbalancer`)
```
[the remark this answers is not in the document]
hausdorff: You mean, it will be hard to convince the various implementors of the `Service` controller to implement a status condition that says something like "this `Service` is fully initialized?" To me it seems like that is the point of having standard conditions, no?

docs/proposals/status.md
barney-s: .reason is not common. We Could add a finalizing condition.

docs/proposals/status.md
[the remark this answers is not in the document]
barney-s: It is. I meant to say it is outside the scope of this proposal. But we can start pushing in that direction after doing it in the core controllers.

ant31:
is the proposal up to date compare to the code that have been merged ? 
I think we should merge it too then.

barney-s:
will sync with code once.

docs/proposals/status.md
barney-s: We will follow up on conditions as an issue

docs/proposals/status.md
barney-s: Ack.

barney-s:
@ant31 @huyhg Please LGTM.
This is synced to the implementation as #81

docs/proposals/status.md
```diff
@@ -0,0 +1,300 @@
...
+3. PodDisruptionBudget is Ready if .`status.currenthealthy >= .status.desiredhealthy`
+... (todo other core resources)
```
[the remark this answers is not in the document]
barney-s: thanks.

barney-s:
/reopen

This is already implemented.

ant31:
@barney-s should this be merged? I think it's important to keep such proposal documentation  in repo

PR #80, opened 2018-11-20
Upgrade to Kubebuilder 1.0.5

mortent:
Upgrades to use Kubebuilder 1.0.5. 

@barney-s
@kow3ns


Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Build the manager binary
+FROM golang:1.10.3 as builder
```
ant31: shouldn't this change be in a separate PR/commit?

mortent: Are you referring to the golang version? This is simply the version used by default in kubebuilder 1.0.5, but we could change it back to what it used to be. I wouldn't think of this change as particularly risky though. But happy to change it if you disagree.

ant31: in general, seems like there re many changes unrelated to upgrading kubebuilder (the commit/pr mentioned).
It on large commit including some configuration changes, documentation, makefile, dockerfile, vendoring updates, code changes...
it make it a bit harder to follow/review and see history later.

mortent:
The critical part of this PR is to make sure that the CRD generated with the latest version of Kubebuilder does not in any significant way differ from the current version. There are some differences, but I don't think they are relevant as they mostly seem to include fields that seem to be missing in the current one or fix some issues where the current CRD and the type didn't match. I have attached the diff here to get some more eyes on it:

diff

barney-s:
LGTM

barney-s:
Could we get this merged @kow3ns @mattfarina  ?

ant31:
The generated CRD looks good, I can't review other parts of the changes.

PR #85, opened 2019-01-25
Add k8s recommended labels

cmoulliard:
Add k8s recommended labels within the code's example of the application operator

### DIFF
--- a/pkg/controller/application/application_controller.go
+++ b/pkg/controller/application/application_controller.go
@@ -44,6 +44,7 @@ const (
 	PartOfLabelKey= "app.kubernetes.io/part-of"
 	ComponentLabelKey= "app.kubernetes.io/component"
 	ManagedByLabelKey= "app.kubernetes.io/managed-by"
+	ManagedBy = "K8s-Application-Operator"
 )
 
 /**
@@ -127,7 +128,14 @@ func (r *ReconcileApplication) Reconcile(request reconcile.Request) (reconcile.R
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{"deployment": instance.Name + "-deployment"},
+				MatchLabels: map[string]string{
+					"deployment": instance.Name + "-deployment",
+					ManagedByLabelKey: ManagedBy,
+					ComponentLabelKey: "Backend",
+					PartOfLabelKey: "Nginx application",
+					NameLabelKey: "nginx-01",
+					VersionLabelKey: "1.0.0",
+				},
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"deployment": instance.Name + "-deployment"}},
--- a/pkg/controller/application/application_controller.go
+++ b/pkg/controller/application/application_controller.go
@@ -129,11 +129,11 @@ func (r *ReconcileApplication) Reconcile(request reconcile.Request) (reconcile.R
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/controller/application/application_controller.go
```diff
@@ -127,7 +128,14 @@ func (r *ReconcileApplication) Reconcile(request reconcile.Request) (reconcile.R
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{"deployment": instance.Name + "-deployment"},
+				MatchLabels: map[string]string{
+					"deployment": instance.Name + "-deployment",
```
ant31: I think we can drop this label

pkg/controller/application/application_controller.go
```diff
@@ -127,7 +128,14 @@ func (r *ReconcileApplication) Reconcile(request reconcile.Request) (reconcile.R
...
+					ManagedByLabelKey: ManagedBy,
+					ComponentLabelKey: "Backend",
+					PartOfLabelKey: "Nginx application",
+					NameLabelKey: "nginx-01",
```
ant31: I would Add 
```
InstanceLabelKey: "01"
```
The `name` could be only "nginx", without 01 (not required)

If a cli/UI has some special usage of those labels it could  look like:
`$AppName ($instance) part of $part-of` 
e.g
`Mysql (production) part of Wordpress`

pkg/controller/application/application_controller.go
cmoulliard: Correct. Will remove it

pkg/controller/application/application_controller.go
cmoulliard: Make sense.

prydonius:
I believe this is just a scaffolded implementation from kubebuilder, not the implementation of the proposed controller (which doesn't intend to create resources like Deployments). Therefore, I'm not sure it makes much sense to have this change, though it doesn't necessarily affect anything.

cmoulliard:
If you don't provided an opinionated scaffolded go file showing how to design an operator supporting Application CRD, then adoption of such technology will be slower as every end user will need to spend time to discover how that should be used. 

Remark: Ideally, the scaffolded code should also include a strategy to update the status of the Application if all the related resources have been created

barney-s:
@cmoulliard #81 removed the scaffolded code and implemented status.
Please take a look at that and rebase on top of it.

cmoulliard:
As you removed the scaffolded code does it make sense that we still add recommended k8s labels as proposed with this PR ? @barney-s

prydonius:
> Is the proposal to inject/propagate the recommended labels in the objects matching the appcrd.selector ?

Is it really the Application controller's job to do this? Shouldn't this be left to the deployment tool or manifest authors? Helm, for example, doesn't inject any labels into resources and expects the chart authors to follow recommended conventions.

cmoulliard:
> Is the proposal to inject/propagate the recommended labels in the objects matching the appcrd.selector ?

If the user has not defined them, then propagating them should be a good practice and specifically if we want to design UI or CLI tools able to fetch the resources part of an application or component

This also what I did for my Component Operator [1] that I will propose to replace the ComponentGroupKind [2] of the application CRD project

[1] https://github.com/snowdrop/component-operator/blob/master/pkg/pipeline/innerloop/install.go#L66
[2] https://github.com/kubernetes-sigs/application/blob/master/pkg/apis/app/v1beta1/application_types.go#L78

cmoulliard:
> If so we need to get these values from a new config the appcrd.spec.recommended_labels or just propagate appcrd.meta.labels

+1 to have a `appcrd.spec.recommended_labels` containing the App recommended labels. As some will be calculated at the level of the component such as `app.kubernetes.io/component`, then such meta should become also part of a Component CRD Spec.

Issue #86, opened 2019-01-31
Proposal: Flag to adopt resources 


barney-s:
Proposal is for adopting resources that match the AppCRD selectors.
```yaml
# Flag that indicates matched resources must be adopted
.spec.adopt: true
```

In some use cases where application crd is used, the tooling explicitly sets OwnerReference of the objects (that match appcrd selectors) to point to the Application object.

Can we make this automatic controller by a flag that is turned off by default.

barney-s:
@kow3ns @huyhg FYI

ant31:
related to https://github.com/kubernetes-sigs/application/issues/7

PR #87, opened 2019-02-02
Add 'addOwnerRef' flag to the spec.

barney-s:
When addOwnerRef is true, inject appcrd as one of the owners for the matching objects
The .controller field of the owner ref is set to false.

fixes https://github.com/kubernetes-sigs/application/issues/86

### DIFF
--- a/pkg/apis/app/v1beta1/application_types.go
+++ b/pkg/apis/app/v1beta1/application_types.go
@@ -84,8 +84,8 @@ type ApplicationSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
-	// Adopt objects - flag to indicate if we need to add OwnerRefs to matching objects
-	Adopt bool `json:"adopt,omitempty"`
+	// AddOwnerRef objects - flag to indicate if we need to add OwnerRefs to matching objects
+	AddOwnerRef bool `json:"addOwnerRef,omitempty"`
 
 	// Info contains human readable key,value pairs for the Application.
 	Info []InfoItem `json:"info,omitempty"`
--- a/pkg/apis/app/v1beta1/application_types.go
+++ b/pkg/apis/app/v1beta1/application_types.go
@@ -84,6 +84,9 @@ type ApplicationSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
+	// Adopt objects - flag to indicate if we need to add OwnerRefs to matching objects
+	Adopt bool `json:"adopt,omitempty"`
+
 	// Info contains human readable key,value pairs for the Application.
 	Info []InfoItem `json:"info,omitempty"`
 
--- a/pkg/apis/app/v1beta1/application_types.go
+++ b/pkg/apis/app/v1beta1/application_types.go
@@ -85,6 +85,7 @@ type ApplicationSpec struct {
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
 	// AddOwnerRef objects - flag to indicate if we need to add OwnerRefs to matching objects
+	// Matching is done by using Selector to query all ComponentGroupKinds
 	AddOwnerRef bool `json:"addOwnerRef,omitempty"`
 
 	// Info contains human readable key,value pairs for the Application.
--- a/pkg/component/types.go
+++ b/pkg/component/types.go
@@ -28,7 +28,7 @@ type Component struct {
 	Handle
 	Name     string
 	CR       metav1.Object
-	OwnerRef []metav1.OwnerReference
+	OwnerRef *metav1.OwnerReference
 }
 
 // KVMap is a map[string]string
--- a/pkg/genericreconciler/genericreconciler.go
+++ b/pkg/genericreconciler/genericreconciler.go
@@ -106,6 +106,9 @@ func specDiffers(o1, o2 metav1.Object) bool {
 	return true
 }
 
+// If both ownerRefs have the same group/kind/name but different uid, that means at least one of them doesn't exist anymore.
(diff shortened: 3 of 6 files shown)


barney-s:
@mortent @huyhg @ant31 @janetkuo 

Please review

pkg/genericreconciler/genericreconciler.go
```diff
@@ -104,6 +106,33 @@ func specDiffers(o1, o2 metav1.Object) bool {
...
+func isReferringSameObject(a, b metav1.OwnerReference) bool {
+	aGV, err := schema.ParseGroupVersion(a.APIVersion)
+	if err != nil {
+		return false
+	}
+	bGV, err := schema.ParseGroupVersion(b.APIVersion)
+	if err != nil {
+		return false
+	}
+	return aGV == bGV && a.Kind == b.Kind && a.Name == b.Name
```
mortent: This check is only sufficient if the two objects is in the same namespace. I think checking the uid rather than the name would be better.

pkg/component/types.go
```diff
@@ -28,7 +28,7 @@ type Component struct {
 	Handle
 	Name     string
 	CR       metav1.Object
-	OwnerRef []metav1.OwnerReference
+	OwnerRef *metav1.OwnerReference
```
mortent: Why not use the slice here? In most situations there can be more than one OwnerReference.

pkg/genericreconciler/genericreconciler.go
barney-s: I like the uid idea.
So does o1.uid == o2.uid => o1.name == o2.name && o1.kind == o2.kind && o1.namespace == o2.namespace ?

OwnerReferences are scoped to namespace and a namespace field doesn't exist in the OwnerReference.

pkg/component/types.go
barney-s: Thats a good question. The object.meta.ownerreferences continue to be a slice. This field refers to the OwnerReference introduced by this controller. The rest of the change is about inserting this controller's reference into the matching objects.

pkg/genericreconciler/genericreconciler.go
```diff
@@ -104,6 +106,33 @@ func specDiffers(o1, o2 metav1.Object) bool {
...
+}
+
+func injectOwnerRefs(o metav1.Object, ref *metav1.OwnerReference) bool {
+	if ref == nil {
+		return false
+	}
+	objRefs := o.GetOwnerReferences()
+	for _, r := range objRefs {
+		if isReferringSameObject(*ref, r) {
```
janetkuo: Following the above comment about checking `uid` or not. If 2 ownerRefs have the same group/kind/name but different uid, that means at least one of them doesn't exist anymore. If we compare `uid` in this function, we'd set both as owners -- probably not what we want, because in the case that the original owner is already gone, we want its dependent to be garbage collected with it, even if a new resource (potential owner) with the same name is created later.

ant31:
other suggestion than 'adopt' ?

- manageOwnerRef
- addOwnerRef
....

pkg/genericreconciler/genericreconciler.go
barney-s: ack.

barney-s:
> * manageOwnerRef
> * addOwnerRef

addOwnerRef seems explicit. I like that.

pkg/genericreconciler/genericreconciler.go
huyhg: It's worth while to document here why we used group/version/kind/name instead of UID. It's hard to understand without the context in which the function is used, similar to Janet's comment below.

@janetkuo I'm aware of the same namespace restriction, but not the cluster scope. Is there documentation somewhere about cluster-scope? FWIW, [kubernetes api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#ownerreference-v1-meta) only mentions namespace.

pkg/apis/app/v1beta1/application_types.go
```diff
@@ -84,6 +84,9 @@ type ApplicationSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
+	// AddOwnerRef objects - flag to indicate if we need to add OwnerRefs to matching objects
+	AddOwnerRef bool `json:"addOwnerRef,omitempty"`
```
ant31: Could you add this field in the documentation / readme too ?

ant31:
does it close # 7?
cc @kow3ns

pkg/apis/app/v1beta1/application_types.go
barney-s: Done. Added.

pkg/genericreconciler/genericreconciler.go
janetkuo: @huyhg I cannot find any document for cluster-scoped owners. The closest I can find is this PR https://github.com/kubernetes/kubernetes/pull/57211 that fixes some bugs about this feature. @caesarxuchao @liggitt we need to document this behavior in https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

janetkuo:
FYI I filed a PR to document the owner ref restriction: https://github.com/kubernetes/website/pull/12543
@barney-s @huyhg @mortent

pkg/apis/app/v1beta1/application_types.go
```diff
@@ -84,6 +84,9 @@ type ApplicationSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 
+	// AddOwnerRef objects - flag to indicate if we need to add OwnerRefs to matching objects
```
janetkuo: Would you document explicitly what "matching objects" mean?

barney-s: done.

barney-s:
@janetkuo added comments describing the ownerref matching.
please lgtm.

Can some one please merge this.

ant31:
/lgtm

need approve

PR #89, opened 2019-02-05
Updated OWNERS file to include link to docs

rlenferink:


rlenferink:
@prydonius thanks for approving. for some reason the approved label is not added. can you retry it with `/approve` ?

Issue #99, opened 2019-03-06
make deploy with right permissions and image


barney-s:
Fix make deploy to use the latest image and with appropriate rbac for the controller.

PR #100, opened 2019-03-07
Adjust roles for deploying application controller

barney-s:
- adjust roles since Application  needs to list and add ownerref to arbitrary resources
- add make undeploy target.

ps: we still need to change the image 

diff --git a/config/default/manager_image_patch.yaml b/config/default/manager_image_patch.yaml                                                                                                                                                
index fcbf39dc..3c7f3e27 100644                                                                                                                                                                                                               
--- a/config/default/manager_image_patch.yaml                                                                                                                                                                                                 
+++ b/config/default/manager_image_patch.yaml                                                                                                                                                                                                 
@@ -8,5 +8,5 @@ spec:                                                                                                                                                                                                                         
     spec:                                                                                                                                                                                                                                    
       containers:                                                                                                                                                                                                                            
       # Change the value of image field below to your controller image URL                                                                                                                                                                   
-      - image: IMAGE_URL                                                                                                                                                                                                                     
+      - image: << My image here >>
         name: manager

### DIFF
--- a/config/rbac/rbac_role.yaml
+++ b/config/rbac/rbac_role.yaml
@@ -5,6 +5,16 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - update
+  - patch
+  - watch
+- apiGroups:
   - app.k8s.io
   resources:
   - '*'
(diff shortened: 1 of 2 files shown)


barney-s:
fixes #99

config/rbac/rbac_role.yaml
```diff
@@ -4,6 +4,12 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
```
ant31: probably only "get, list, update, patch, watch"

barney-s: ack

PR #109, opened 2019-06-14
Enable status subresource, update manifest

halfcrazy:


barney-s:
Thanks for the change. 
Would you mind adding a simple e2e test here to ensure the status endpoint works.
https://github.com/kubernetes-sigs/application/blob/master/e2e/main_test.go#L88

barney-s:
#145 adds status.

/close

Issue #115, opened 2019-10-10
ConfigMaps from application deleted a second after creation


barney-s:
related to https://github.com/GoogleCloudPlatform/click-to-deploy/issues/592

Issue:
- statefulsets creation fails due to missing configmaps
- configmaps from application are created successfully by the deployer and deleted by application-controller-manager a second later;

Logs:

```
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)  Expected Resources:
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.ConfigMap/prometheus-1-deployer-config
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.Deployment/prometheus-1-kube-state-metrics
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1beta2.Deployment/prometheus-1-kube-state-metrics
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1beta1.Deployment/prometheus-1-kube-state-metrics
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.PersistentVolumeClaim/prometheus-1-alertmanager-data-prometheus-1-alertmanager-0
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.PersistentVolumeClaim/prometheus-1-grafana-data-prometheus-1-grafana-0
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.Secret/prometheus-1-grafana
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.Service/prometheus-1-grafana
2019/10/08 16:24:15 *v1beta1.Application/prometheus-1/prometheus-1(cmpnt:app)   exp: prometheus-1/*v1.Service/prometheus-1-kube-state-metrics
[...]

PR #116, opened 2019-10-10
Avoid observe after mutating

barney-s:
Problem:
related to GoogleCloudPlatform/click-to-deploy#592

Issue:
1. An application was deployed from market place - application crd, sts, configmaps, ...
2. The application controller deleted a config map a few seconds later
3. This was an intermittent problem

Root Cause:
The ObserverAndMutate() code does this:
1. observed <- observe resources from cluster
2. expected <- controller's mutate(observed) [in this controller, expected = observed]
3. observed <- observe resources from cluster
4. reconcile expected and observed => observed between 1, 3 differs thereby differs from expected.
  
As part of reconcile we delete objects that are in observed but not in expected.

Solution:
The logic to observe after a mutation does not apply to this controller. It was from a different code.
We can safely remove the second observe()

### DIFF
--- a/pkg/genericreconciler/genericreconciler.go
+++ b/pkg/genericreconciler/genericreconciler.go
@@ -219,15 +219,6 @@ func (gr *Reconciler) ObserveAndMutate(crname string, c component.Component, sta
 		if err != nil {
 			return emptybag, emptybag, fmt.Errorf("Failed mutating resources: %s", err.Error())
 		}
-
-		// Get observables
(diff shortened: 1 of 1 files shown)


barney-s:
@mortent @janetkuo Please review

barney-s:
Thanks morten. This fixes https://github.com/kubernetes-sigs/application/issues/115

barney-s:
The test case would be make sure nothing is deleted which is something of a hard-requirement :)

barney-s:
@mortent  @kow3ns  @liyinan926  Could one of you LGTM the change ?

PR #119, opened 2019-10-22
Generate client for application type

richardsliu:
Also added an update-codegen script.


janetkuo:
The client was deleted in #80 when we upgrade kubebuilder version. 

To get a client from kubebuilder - https://github.com/kubernetes-sigs/application/blob/master/README.md#programmatically

@mortent

PR #124, opened 2019-11-01
sort the status componentlist so that it converges stably

barney-s:
Sort the componentlist objects in a stable way so that the status converges.
this avoids the controller from looping because it observes changes in application cr since the order changes


barney-s:
The 

> `Conditions` in `ApplicationStatus` is also a slice. Is that sorted too?

The conditions are built from code in a specific order. So they always converge in the same order.

barney-s:
@mortent  @kow3ns  @liyinan926  Could one of you LGTM the change ?

PR #127, opened 2019-11-11
add option to run controller against a namespace

IronPan:
Context: there are some user, especially enterprise customers who wants to use application CRD but don't want to install ClusterRole/ClusterRoleBinding since it has cluster scoped permission. In this case the cluster is a shared multi-user cluster, and cluster admin wants to scope the application controller to scan only the namespace a user owns. 

This change adds a mode to run controller against only a certain namespace.


barney-s:
Please resolve the conflicts (rebase) and ping again.

main.go
```diff
@@ -44,8 +44,10 @@ func init() {
 }
 
 func main() {
+	var namespace string
 	var metricsAddr string
 	var enableLeaderElection bool
+	flag.StringVar(&namespace, "namespace", "kubeflow", "Namespace within which CRD controller is running.")
```
barney-s: please remove kubeflow. Let the default be ""

Dockerfile
```diff
@@ -45,4 +45,5 @@ FROM gcr.io/distroless/static:latest
 WORKDIR /
 COPY --from=builder /workspace/manager .
 USER nobody
-ENTRYPOINT ["/manager"]
+ENV NAMESPACE ""
+ENTRYPOINT ["/manager", "--namespace=${NAMESPACE}"]
```
barney-s: Please remove the parameter option here. 
Pass it via:
https://github.com/kubernetes-sigs/application/blob/master/config/manager/manager.yaml#L46

main.go
IronPan: done

Dockerfile
IronPan: done. and i think this file can be configured as needed.

Issue #137, opened 2020-01-29
e2e test flake: Application CRD should install correctly


janetkuo:
Test failure from #128:
https://travis-ci.org/kubernetes-sigs/application/jobs/643342852

```
=== RUN   TestE2e
Running Suite: Application Type Suite
=====================================
Random Seed: 1580299211
Will run 4 of 4 specs
•Creating new Application
------------------------------
• Failure [0.039 seconds]
Application CRD should install correctly
/home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:64
  should register an application [It]
  /home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:90
  Expected error:
      <*meta.NoKindMatchError | 0xc0002d0b00>: {
          GroupKind: {Group: "app.k8s.io", Kind: "Application"},
          SearchedVersions: ["v1beta1"],
      }
      no matches for kind "Application" in version "app.k8s.io/v1beta1"
  not to have occurred
  /home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:93
------------------------------
• Failure [0.095 seconds]
Application CRD should install correctly
/home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:64
  should delete application [It]
  /home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:96
  Expected error:
      <*errors.StatusError | 0xc0001443f0>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
              Status: "Failure",
              Message: "applications.app.k8s.io \"wordpress-01\" not found",
              Reason: "NotFound",
              Details: {Name: "wordpress-01", Group: "app.k8s.io", Kind: "applications", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 404,
          },
      }
      applications.app.k8s.io "wordpress-01" not found
  not to have occurred
  /home/travis/gopath/src/github.com/kubernetes-sigs/application/e2e/main_test.go:99
------------------------------
•Failed to create JUnit report file: /workspace/_artifacts/junit.xml
	open /workspace/_artifacts/junit.xml: no such file or directoryFailed to generate JUnit report
	invalid argument
Summarizing 2 Failures:
[Fail] Application CRD should install correctly [It] should register an application 
[...]

PR #138, opened 2020-01-30
Stablize the E2E test

nan-yu:


nan-yu:
Fixes: https://github.com/kubernetes-sigs/application/issues/137

barney-s:
Just to confirm we are removing prow since we dont want it for this repo. Just travis.

Why are you deleting the nodes as well from kind ?
We may need them once we extend e2e tests to include status tests.
May be have 1 node instead of 3.

Issue #102, opened 2019-03-23
Support GroupVersion and Groups in spec.componentKinds 


barney-s:
The proposal is to support both GroupVersion and Group in the spec.componentKinds. 
We will ignore and not validate the version part of the GroupVersion.
The field type and name will still reflect GroupKind to guide the user. But we are relaxing the allowed values to reduce user friction.

PR #139, opened 2020-01-31
Migrate application CRD to kubebuilder 2.0

nan-yu:


controllers/status.go
```diff
@@ -0,0 +1,256 @@
+package controllers
```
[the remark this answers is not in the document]
nan-yu: I did use `git mv` to move the file from the older directory to new directory (see below). However, after applying the changes, `git` doesn't think it is a `git mv` action as the code in `status.go` changed dramatically.

An alternative way is to split it into 2 commits.
The first commit will just be a simple `git mv`, and the second commit will apply the update.
The downside of this way is that the first commit will not compile.

I would prefer to keep it as is. Let me know if you have any objections. @barney-s 

```
✔ ~/go/src/sigs.k8s.io/application [master ↑·1|…1⚑ 2] 
10:13 $ git mv pkg/apis/app/v1beta1/condition.go controllers/
✔ ~/go/src/sigs.k8s.io/application [master ↑·1|●1…1⚑ 2] 
10:13 $ git mv pkg/apis/app/v1beta1/status.go controllers/
✔ ~/go/src/sigs.k8s.io/application [master ↑·1|●2…1⚑ 2] 
10:13 $ gst
On branch master
Your branch is ahead of 'upstream/master' by 1 commit.
  (use "git push" to publish your local commits)

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    pkg/apis/app/v1beta1/condition.go -> controllers/condition.go
	renamed:    pkg/apis/app/v1beta1/status.go -> controllers/status.go


```

controllers/conditions.go
```diff
@@ -0,0 +1,72 @@
+package controllers
+
```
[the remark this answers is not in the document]
nan-yu: ditto

nan-yu:
@barney-s comments resolved. I did `git commit --amend` to update the commit. Let me know if you want me to push new commits for later update. Thanks!

controllers/application_controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	newApplicationStatus := r.getNewApplicationStatus(ctx, &app)
+
+	if equality.Semantic.DeepEqual(newApplicationStatus, &app.Status) {
+		return ctrl.Result{}, nil
+	}
+
+	newApplicationStatus.ObservedGeneration = app.Generation
+	if err = r.updateApplicationStatus(ctx, req.NamespacedName, newApplicationStatus); err != nil {
+		return ctrl.Result{Requeue: true}, err
```
mortent: I don't think you need to set Requeue here. If the error is not nil, the controller library will requeue as needed.

controllers/application_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	newApplicationStatus := r.getNewApplicationStatus(ctx, &app)
+
+	if equality.Semantic.DeepEqual(newApplicationStatus, &app.Status) {
+		return ctrl.Result{}, nil
+	}
+
+	newApplicationStatus.ObservedGeneration = app.Generation
```
mortent: Should we do this before we check whether there are any changes to the status and possibly return early? Seems like the application has been observed at the given version even if there are no changes to the status.

controllers/application_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		if err := r.setOwnerRefForResources(ctx, *ownerRef, resources); err != nil {
+			return resources, err
+		}
+	}
+	return resources, nil
+}
+
+func (r *ApplicationReconciler) getNewApplicationStatus(ctx context.Context, app *appv1beta1.Application) *appv1beta1.ApplicationStatus {
+
+	resources, err := r.updateComponents(ctx, app)
```
mortent: I think it would be cleaner if we kept updating the owner refs separate from computing the status. The first includes updates to resources not managed by the application controller (so relatively risky operations), while the second is just computing their status. Currently updating the owner refs looks like a side effect of computing the new application status.

controllers/application_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	return newApplicationStatus
+}
+
+func (r *ApplicationReconciler) fetchComponentListResources(ctx context.Context, groupKinds []metav1.GroupKind, selector *metav1.LabelSelector, namespace string) ([]*unstructured.Unstructured, error) {
+	logger := getLoggerOrDie(ctx)
+	var resources []*unstructured.Unstructured
+	for _, gk := range groupKinds {
+		mapping, err := r.Mapper.RESTMapping(schema.GroupKind{
```
mortent: In this code, we should make sure we can handle situations where the Group in the list of GroupKinds from the App CR actually contains the GroupVersion. According to the spec it should only contain the group, but I know there are places where the group is included and I think this is handled by the old version of the controller.

controllers/status.go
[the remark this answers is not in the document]
barney-s: No Objects. Thanks.

controllers/conditions.go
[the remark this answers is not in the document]
barney-s: ack

controllers/application_controller.go
nan-yu: If we move this before we check whether there are changes to the status, the observedGeneration field will keep increasing. I'll take a deep look into why it is happening. Before we figure it out, I suggest we keep it as is to make sure we get correct observedGeneration field.

controllers/application_controller.go
nan-yu: DONE with the `StripVersion` function. Tests added for such case: https://github.com/kubernetes-sigs/application/blob/1abaa21daac7a901f17ce2049c1270cd5ebfe18a/controllers/application_controller_test.go#L171

barney-s: Right StripVersion was added to address https://github.com/kubernetes-sigs/application/issues/102

Dockerfile.ppc64le
```diff
@@ -1,17 +1,27 @@
 # Build the manager binary
```
barney-s: Please remove this file. Since the Dockerfile is parameterized for multi-arch builds.

Makefile
```diff
@@ -68,54 +71,57 @@ all: test manager
...
-	kubectl apply -k config/crds
+	kubectl apply -k config/crd
+
+# Uninstall CRDs from a cluster
+uninstall: $(KUSTOMIZE)
+	kubectl delete -k config/crd
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: $(KUSTOMIZE)
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
```
barney-s: PLease add back the kustomize edit here as well. Something similar to this:
cd config/manager && kustomize edit set image controller=${IMG}

Makefile
```diff
@@ -138,23 +146,22 @@ generate: ## Generate code
...
 ## Docker
 ## --------------------------------------
 
 .PHONY: docker-build
-docker-build: ## Build the docker image for controller-manager
+docker-build: test ## Build the docker image for controller-manager
 	docker build --network=host --pull --build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
 	@echo "updating kustomize image patch file for manager resource"
-	sed -i'' -e 's@image: .*@image: '"${CONTROLLER_IMG}-$(ARCH):$(TAG)"'@' ./config/default/manager_image_patch.yaml
+	sed -i'' -e 's@image: .*@image: '"${CONTROLLER_IMG}-$(ARCH):$(TAG)"'@' ./config/manager/manager.yaml
```
barney-s: Use kustomize edit above instead of this.

go.mod
```diff
@@ -3,7 +3,9 @@ module sigs.k8s.io/application
 go 1.13
 
 require (
-	github.com/golang/groupcache v0.0.0-20181024230925-c65c006176ff // indirect
+	github.com/go-logr/logr v0.1.0
+  github.com/golang/groupcache v0.0.0-20181024230925-c65c006176ff // indirect
```
barney-s: nit: align with other lines please.

PR #140, opened 2020-02-02
Proposal for inventory object

barney-s:
Opening a proposal for inventory object for further discussion,

### DIFF
--- a/enhancements/inventory-object.md
+++ b/enhancements/inventory-object.md
@@ -1,0 +1,64 @@
+# Inventory Object Proposal
+
+## Motivation
+The proposal is to seperate the Application CRD's object selector and group kinds into a separate object. Separating the resource grouping and Application info allows us to reuse resoruce grouping (inventory object) in other controllers and tools. The Inventory object is a descriptive object and needs no controller and status field.
+
+## Summary
+
+- Introduce an Inventory object and a data format that is used to group objects
+- Allow selector and groupKind based grouping
+- Allow explicit object list based grouping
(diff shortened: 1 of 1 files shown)


enhancements/inventory-object.md
```diff
@@ -0,0 +1,64 @@
...
+## Summary
+
+- Introduce an Inventory object and a data format that is used to group objects
+- Allow selector and groupKind based grouping
+- Allow explicit object list based grouping
+- Reference array of Inventory objects in Application object
+
+## Proposal
+
+We propose to introduce a separate Inventory/Grouping object. It could be a new API Type or use ConfigMap to achieve the same.
```
ant31: Do you plan to create an 'instance' of the inventory ? If not how configmap could be an alternative ?

ant31:
Do you plan to make it only an internal type to application object, or also a create an instance of the inventory in the cluster?

barney-s:
The proposal is to either use ConfigMap (to align with https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/prune/grouping.go#L165) Or add a new CRD Inventory which has its own spec and no status.

ant31:
Should we have a meeting about it? It's very unclear for me, but one of the goal is to get lire closely integrated with kustomize and it seems a path forward.
could add it on the sigcli or sigapps meeting agenda?

mattfarina:
@barney-s what would this look like from an end-user perspective? It adds complexity and I was wondering what the experience would be.

PR #145, opened 2020-02-06
Add the status subresource to the application CRD

nan-yu:
The status subresource was not added to the CRD definition, which makes
`r.Client.Status().Update(ctx, original)` return a 404 error with the
following message:
```
failed to update status of Application default/xxx:
applications.app.k8s.io "xxx" not found
```

This commit add the status subresource to the CRD definition.

It also moves application observedGeneration before we check whether
there are any changs to the status, which makes it possibly return
early.

### DIFF
--- a/config/crd/bases/app.k8s.io_applications.yaml
+++ b/config/crd/bases/app.k8s.io_applications.yaml
@@ -15,6 +15,8 @@ spec:
     plural: applications
     singular: application
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       description: Application is the Schema for the applications API
@@ -471,6 +473,8 @@ spec:
               format: int64
               type: integer
           type: object
+      required:
+      - spec
       type: object
   version: v1beta1
   versions:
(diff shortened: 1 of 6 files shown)


barney-s:
Is there a test which captures this ?
Check for status end point ?

nan-yu:
The e2e test should capture this change, but given the e2e test is not automated yet, I'm going to add a unit test for the endpoint.

config/crd/bases/app.k8s.io_applications.yaml
```diff
@@ -471,7 +473,8 @@ spec:
               format: int64
               type: integer
           type: object
-      type: object
+      required:
```
nan-yu: I have to remove `type: object`, otherwise the test would fail with the error: `must only have "properties",  "required" or "description" at the root if the status subresource is enabled occurred`

nan-yu: The above validation was from an older version of kubernetes: v1.11.0. I'm going to update k8s to 1.16.4 in the test. That should fix the validation failure without removing `type: object`.

mortent: I think that sounds ok. In general we should make sure it runs on the last 3-4 versions of Kubernetes. Clusters with older versions than that are no longer supported by the community, so we can forget about them.

Issue #147, opened 2020-02-07
Error when application resource deleted ERROR recorded in logs


lewismarshall:
**Using local build of master@01d8bf78**

Deleting an application resource will produce an error in the logs after objects can't be found until controller restarted:

```
2020-02-07T10:44:20.186Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "application", "request": "application-system/captain", "error": "Application.app.k8s.io \"captain\" not found"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.16.5/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.16.5/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
	/go/pkg/mod/k8s.io/apimachinery@v0.16.5/pkg/util/wait/wait.go:88
```

The application resource is here:
```
apiVersion: app.k8s.io/v1beta1
kind: Application
metadata:
  name: "captain"
  labels:
    kore-direct: "true"
spec:
  selector:
    matchLabels:
      app: captain
      service_name: captain
  componentKinds:
    - group: controller
      kind: Service
    - group: controller
      kind: Deployment
  assemblyPhase: "Pending"
  descriptor:
    version: ""
    description: "A controller for deploying helm3 charts"
    type: "captain"
```

barney-s:
@lewismarshall Just to confirm is this with latest code base ?

[...]

PR #149, opened 2020-02-07
Fix the not found ERROR in the log when app deleted

nan-yu:


nan-yu:
fix: https://github.com/kubernetes-sigs/application/issues/147

Issue #131, opened 2019-12-16
Add support for protocol in service ref 


barney-s:
Add support for protocol string in the service and ingress info sections.
This helps with users of application crd to construct URL with appropriate protocol.

https://EXTERNAL-IP
or
http://EXTERNAL-IP

https://github.com/kubernetes-sigs/application/issues/95

Issue #95, opened 2019-02-22
ingressRef does not support HTTPS protocol


wgrzelak:
Currently `ingressRef` supports only HTTP protocol. I would like to have there HTTPS protocol as well.

Current state:

```yaml
 - name: WordPress site address
    type: Reference
    valueFrom:
      ingressRef:
        name: $APP_INSTANCE_NAME-wordpress-ingress
```

Preferred state:

```yaml
 - name: WordPress site address
    type: Reference
    valueFrom:
      ingressRef:
        name: $APP_INSTANCE_NAME-wordpress-ingress
        protocol: HTTPS
```

I think the same needs could be with `serviceRef` as well.

prydonius:
@wgrzelak can't the protocol be worked out by the client by requesting the Ingress object and reading the host and tls configuration in that object?

PR #150, opened 2020-02-07
Add protocol fields to Service and Ingress info items

barney-s:
For #131


barney-s:
Please review @nan-yu @huyhg

barney-s:
@mortent please review

huyhg:
LGTM

mortent:
I don't see any big issues with this change, but could you explain why we are adding this? Either in the PR or in the issue.

barney-s:
It was a request from an user of application CR #95 

Add support for protocol string in the service and ingress info sections.
This helps with users of application crd to construct URL with appropriate protocol.

https://EXTERNAL-IP
or
http://EXTERNAL-IP

PR #155, opened 2020-02-10
Grant application-manager access to all resources

nan-yu:
The application controller will list and update resources.
Without the access, the controller will fail to fetch or update the
resources with the following errors:
```
controllers.Application	deployments.apps is forbidden: User "system:serviceaccount:application-system:default" cannot list resource "deployments" in API group "apps" in the namespace "default"
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -29,3 +29,10 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - list
+  - update


config/rbac/role.yaml
```diff
@@ -29,3 +29,10 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - list
+  - update
```
barney-s: I dont think we need update access for all apiGroups and resources.

nan-yu: I think we do. Otherwise, we have to list all known GVKs. Also, for future CRDs, we will get the same error when we're trying to list/update them.

barney-s: i see update for OwnerRef injection.

ant31: this is a generated file @nan-yu and will be overwritten

ant31: I created a PR #174

Issue #112, opened 2019-09-03
What is the use of ConfigMapKeySelector? How to get ConfigMap in app get?


hemantha-kumara:
Can you explain what is the use of ConfigMapKeySelector. Is it used by the Application to list the configmaps that are used by it?

When an application is created and "ConfigMap" is included in the application definition, the output of 

> kubectl get application 

does not list the configMap. I have looked into the issue [ConfigMap not listed](https://github.com/kubernetes-sigs/application/issues/104) where the user had an incorrect Group. The Group is correctly set to core in my case.

Are these two related?
@nrchakradhar

barney-s:
Try using 

```
group: v1
kind: ConfigMap
```

janetkuo:
Ref #113

Please try again using `v1` (instead of `core`) as group. If it still doesn't work, please reopen this issue. 

cc @nan-yu for doc improvement

nan-yu:
sample updated in PR https://github.com/kubernetes-sigs/application/pull/156

PR #156, opened 2020-02-10
Update the group of Service in the examples and tests

nan-yu:
The group of Service can be either "v1" or empty (""), but not "core",
so as other resources (ConfigMap, Pod, Secret, PersistentVolumeClaim and
etc.).

### DIFF
--- a/config/samples/app_v1beta1_application.yaml
+++ b/config/samples/app_v1beta1_application.yaml
@@ -13,7 +13,7 @@ spec:
     matchLabels:
       app.kubernetes.io/name: "wordpress-01"
   componentKinds:
-  - group: core
+  - group: v1
     kind: Service
   - group: apps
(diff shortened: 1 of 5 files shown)


config/samples/app_v1beta1_application.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     matchLabels:
       app.kubernetes.io/name: "wordpress-01"
   componentKinds:
-  - group: core
+  - group: v1
```
barney-s: I understand this change. Why change "" to v1 in the others ?

nan-yu: per Janet's comment: https://github.com/kubernetes-sigs/application/issues/112#issuecomment-580008476.
It seems `v1` makes more sense than `""`.

PR #157, opened 2020-02-10
use kstatus

barney-s:
Use kstatus from kustomize (lineage from controller-reconciler, application, cli-utils)
Fix lint issues
Run license after manifests


barney-s:
@nan-yu Please review

mortent:
We have moved this code to cli-utils and the code in the kustomize repo will be deleted. We don't have any releases from cli-utils yet, so I suggest we hold off on make this change for a bit.

barney-s:
> We have moved this code to cli-utils and the code in the kustomize repo will be deleted. We don't have any releases from cli-utils yet, so I suggest we hold off on make this change for a bit.

Do we have an ETA ? If it is going to be O(month) we can use this for now and then move to the cli-utils one. This would give us crucial time to soak the change as part of existing effort to qualify.

mortent:
I would think we have something in a couple of weeks. But could we just keep the current implementation for now, get the current changes rolled out and then address this in a follow-up? 

Seems like we already have a lot of changes going in for the next release, so maybe we should hold off on this? The changes so far should not change any of the behavior of the controller, so it should be a simple drop-in replacement. Changing the rules we use to compute status means we are changing the behavior.

barney-s:
Sounds fair. We can wait on this merge till after release

PR #158, opened 2020-02-12
Add a more comprehensive example

nan-yu:
The new example includes apps.Deployment, batch.Job, Service, ConfigMap and a custom CRD.


nan-yu:
/assign @barney-s

I only added a bash script for deployment, but not for test, because the e2e test will be almost the same as wordpress.

docs/examples/test-app/test_crd.yaml
```diff
@@ -0,0 +1,56 @@
...
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
```
barney-s: Could we add a condition struct array here ? 

Similar to 
https://github.com/kubernetes-sigs/application/blob/master/config/crd/bases/app.k8s.io_applications.yaml#L445

docs/examples/test-app/deploy.sh
```diff
@@ -0,0 +1,30 @@
...
+set -o nounset
+set -o pipefail
+
+# Get a Kubernetes cluster ready before deploying the application
+
+# Deploy the Application CRD with the controller on the Kubernetes cluster
+make deploy
+
+# Install the test CRD to the cluster
+kubectl apply -f test_crd.yaml
```
barney-s: We may have to wait for the crd to be registered.

docs/examples/test-app/deploy.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Create the application object
+kubectl apply -f application.yaml
+
+# Verify application's status/components field shows all components
+# Verify ownerReferences is added to all components' spec
```
barney-s: Does the script wait by itself ?
Do we need a readln OR a sleep here ?

docs/examples/test-app/test_crd.yaml
nan-yu: This test includes CRD only without the controller, so the conditions wouldn't be updated and it makes no difference if we have a condition struct array or not. For simplicity, I would keep it empty.

docs/examples/test-app/deploy.sh
nan-yu: We validate the components' specs and application's status manually, but I can add a readln in the script just in case we may want to automate the script later.

docs/examples/test-app/deploy.sh
nan-yu: I will add a readln to confirm the CRD is registered.

barney-s:
please take a look at #159 
We could automate this test once that is merged. 
Holding this till then.

nan-yu:
awesome! Do we want to add a new e2e test for this comprehensive case? Or do we keep the bash script for now?

docs/examples/test-app/application.yaml
```diff
@@ -0,0 +1,123 @@
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: "test-application-01"
+  labels:
+    app.kubernetes.io/name: "test-01"
+spec:
```
ant31: Let's encourage to use descriptor, especially once it's integrated with kubectl/kubernetes-dashboard that would be the first experience user get and missing fields wouldn't look good on those tools.

nan-yu:
@barney-s ready to be re-reviewed and merged

barney-s:
Thanks for updating. 
Would you think using a kustomize.yaml + individual resources would be more readable ? 
That way we can have 2 kustomize, 1 for good case and 1 for bad case with the same set of manifests.

If you want to do a follow up PR also it ok.
Let me know.

barney-s:
/lgtm
/approve 

we can reorg the files in a follow up PR
Functionally this is good.

PR #159, opened 2020-02-12
E2E test improvements

barney-s:
E2E test improvements

- Run the controller in e2e test
- Combine wordpress test with the existing e2e test
- Use a kind cluster to do real e2e
- Make example author and developer generic


barney-s:
@nan-yu - Please review. This runs the controller and does an actual E2E test .

nan-yu:
LGTM. Thanks for automating the e2e test. It's much clean now!!! @barney-s

barney-s:
@mortent - Would you please review and LGTM

main.go
```diff
@@ -33,9 +33,11 @@ func init() {
 func main() {
 	var namespace string
 	var metricsAddr string
+	var syncPeriod int
 	var enableLeaderElection bool
 	flag.StringVar(&namespace, "namespace", "", "Namespace within which CRD controller is running.")
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.IntVar(&syncPeriod, "sync-period", int(2*time.Minute), "Sync every sync-period seconds.")
```
nan-yu: int(2*time.Minute) is sync-period **nanoseconds**, not seconds.
We need to convert nanoseconds to seconds or update the flag description.

barney-s: Thanks. Fixed it.

main.go
```diff
@@ -44,13 +46,13 @@ func main() {
 		o.Development = true
 	}))
 
-	syncPeriod := 2 * time.Minute
+	syncPeriodD := time.Duration(int(time.Second) * syncPeriod)
```
nan-yu: That might overflow on different machines.

I used the `36000` (10hours) to play with it. It was running okay on my glinux, but overflow in golang playground: https://play.golang.org/
```
package main

import (
	"fmt"
	"time"
)

func main() {
        var syncPeriod int = 36000
	syncPeriodD := time.Duration(int(time.Second) * syncPeriod)
	fmt.Println("syncPeriodD = ", syncPeriodD)
}
```

I got `syncPeriodD =  -415.875072ms`. I think we need to take care of it.

barney-s: I would think the playground is a 32 bit env.
explicitly using int64 works. 

func main() {
        var syncPeriod int64 = 36000
	syncPeriodD := time.Duration(int64(time.Second) * syncPeriod)
	fmt.Println("syncPeriodD = ", syncPeriodD)
}

barney-s: Fixed to use int64

barney-s:
@ant31 @janetkuo  - would you please review and LGTM ?

config/rbac/role.yaml
```diff
@@ -29,10 +29,3 @@ rules:
   - get
   - patch
   - update
```
nan-yu: Are these removed by accident?

barney-s: looks like it. curious why it is generated this way. will remove it.

nan-yu: I meant we should add them back. They were added via PR: https://github.com/kubernetes-sigs/application/pull/155.

barney-s: restored.

Issue #148, opened 2020-02-07
Condition Ready is True when Deployment replicas are 0


lewismarshall:
Using - master@01d8bf78

It would seem as though the controller isn't updating the status when deployment scaled (or even if a deployment deleted and re-created with bad image):
```
$ kubectl -n captain-system scale deploy/captain-controller-manager --replicas=0
deployment.extensions/captain-controller-manager scaled
$ date ;kubectl -n captain-system get deploy,svc -l app=captain -o wide
Fri  7 Feb 2020 12:34:34 GMT
NAME                                               READY   UP-TO-DATE   AVAILABLE   AGE    CONTAINERS   IMAGES                                 SELECTOR
deployment.extensions/captain-controller-manager   0/0     0            0           5d1h   manager      index.alauda.cn/claas/captain:v0.9.4   app=captain

NAME                      TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)            AGE    SELECTOR
service/captain-webhook   ClusterIP   10.0.0.181   <none>        443/TCP,6060/TCP   5d1h   app=captain
```
And then the pertinent logs after the next reconciliation period:
```
$kubectl -n application-system logs -l control-plane=application-manager -c manager
...
2020-02-07T12:36:31.857Z	INFO	controllers.Application	NoMappingForGK	{"application": "captain-system/captain", "gk": "Service.controller"}
2020-02-07T12:36:33.910Z	INFO	controllers.Application	NoMappingForGK	{"application": "captain-system/captain", "gk": "Deployment.controller"}
2020-02-07T12:36:33.910Z	DEBUG	controller-runtime.controller	Successfully Reconciled	{"controller": "application", "request": "captain-system/captain"}
...
```
And finally the application object after the reconciliation:
```
$ date ; kubectl -n captain-system get application -o yaml
Fri  7 Feb 2020 12:37:37 GMT
apiVersion: v1
items:
- apiVersion: app.k8s.io/v1beta1
  kind: Application
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #167, opened 2020-02-14
Update the way to compute resource status

nan-yu:
This commit follows the documentation to compute the resource status:
https://docs.google.com/document/d/1RAPC5-LCb_VBpzvbpCLn8Z-7eilRAWmO0RGbVuChhPc/edit#heading=h.rqs3jxocuzbw

Instead of extracting fields from unstructured, it converts unstructured to corresponding object first.

### DIFF
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -4,12 +4,14 @@
 package controllers
 
 import (
-	"fmt"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // Constants defining labels
@@ -38,6 +40,10 @@ func status(u *unstructured.Unstructured) (string, error) {
 		return podStatus(u)
 	case "PodDisruptionBudget.policy":
 		return pdbStatus(u)
+	case "ReplicationController":
+		return replicationControllerStatus(u)
+	case "Job.batch":
+		return jobStatus(u)
 	default:
 		return statusFromStandardConditions(u)
 	}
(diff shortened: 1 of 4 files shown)


nan-yu:
@barney-s @mortent I found this [public doc](https://docs.google.com/document/d/1RAPC5-LCb_VBpzvbpCLn8Z-7eilRAWmO0RGbVuChhPc/edit#heading=h.rqs3jxocuzbw) online and made the change accordingly.
Please take a look to see if it makes sense to you. If it looks okay, I can add more test around it.

nan-yu:
The second commit is to fix the error mentioned in https://github.com/kubernetes-sigs/application/issues/148

In such case, the Ready condition should be "False".

```
status:
  conditions:
  - lastTransitionTime: "2020-02-13T16:30:39Z"
    lastUpdateTime: "2020-02-13T16:30:39Z"
    message: all components ready
    reason: ComponentsReady
    status: "True"
    type: Ready
  - lastTransitionTime: "2020-02-13T16:30:39Z"
    lastUpdateTime: "2020-02-13T16:30:39Z"
    message: 'deployments.apps is forbidden: User "system:serviceaccount:application-system:default"
      cannot list resource "deployments" in API group "apps" in the namespace "kore-system"'
    reason: ErrorSeen
    status: "True"
    type: Error
```

barney-s:
This may qualify as change in current behavior. Let's cut a release and then immediately merge this for the next release. @mortent  ?

nan-yu:
@barney-s @mortent I'm inclined to include this change in the release as this is not an enhancement, but more like bug-fix to me. Otherwise, we have to cut another release.

barney-s:
Would you please squash the commits ?

https://github.com/wprig/wprig/wiki/How-to-squash-commits

nan-yu:
> Would you please squash the commits ?
I was explicitly making it 3 separate commits, each with a dedicated purpose. Let me know if squash is required before merging, otherwise, I'm leaning towards keeping it as is. @barney-s

barney-s:
Could we add Job and CronJob as well to this list ?
Refer to this:
https://github.com/kubernetes-sigs/cli-experimental/blob/master/internal/pkg/status/legacy_status.go

barney-s:
You could still squash the commits and add a multiline comment.

nan-yu:
CronJob will always be ready and fall into the default case.
For Job, it seems that job is not-ready only when job.spec.startTime is not set
```
func jobStatus(u *unstructured.Unstructured) (string, error) {
	job := &batchv1.Job{}

	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, job); err != nil {
		return "", err
	}

	if job.Status.StartTime == nil {
		return StatusInProgress, nil
	}

	return StatusReady, nil
}
```

Commits are squashed.

barney-s:
Thanks @nan-yu 
Will review and lgtm

controllers/application_controller.go
```diff
@@ -87,22 +87,31 @@ func (r *ApplicationReconciler) updateComponents(ctx context.Context, app *appv1
...
 		setReadyCondition(newApplicationStatus, "ComponentsReady", "all components ready")
 	} else {
 		setNotReadyCondition(newApplicationStatus, "ComponentsNotReady", "some components not ready")
 	}
 
-	if err != nil {
-		setErrorCondition(newApplicationStatus, "ErrorSeen", err.Error())
+	if len(errs) != 0 {
+		for _, err := range errs {
+			setErrorCondition(newApplicationStatus, "ErrorSeen", err.Error())
```
barney-s: This will keep overwriting the error condition.
Maybe consider aggregating the errors.

controllers/application_controller.go
```diff
@@ -179,15 +187,12 @@ func (r *ApplicationReconciler) objectStatuses(ctx context.Context, resources []
 		}
 		s, err := status(resource)
 		if err != nil {
-			// Just logging the error for now. Not sure if this is the right way to handle it.
-			logger.Error(err, "unable to compute status for resource", "gvk", resource.GroupVersionKind().String(),
```
barney-s: Is logging not needed ?

barney-s: Also there is a reason why we continue and not bail out. 
A single bad object would cause us from not aggregating status from other objects. It was a bad UX.

controllers/application_controller.go
nan-yu: Will aggregate the errors in a separate PR. This PR is aiming to only change the way how do we compute status.

controllers/application_controller.go
nan-yu: Will add logging back and continue compute status from other resources even if error occurs.
Will do in a separate PR to keep this one clean.

barney-s:
/approve
/lgtm

we may need status_test.go at some point.

PR #160, opened 2020-02-13
Update docs

barney-s:
Update docs

- Add development guide
- Add quickstart guide
- Simplify Readme
- Move api table from readme to api.md
- Update contributing.md

PR #169, opened 2020-02-15
Fix doc broken links

barney-s:
Fix review comments from https://github.com/kubernetes-sigs/application/pull/160


barney-s:
@nan-yu  - thanks for the review comments. I have fixed those.

Issue #136, opened 2020-01-28
[Question] CRD in componentKinds?


tzununbekov:
Repository [readme](https://github.com/kubernetes-sigs/application/blob/88337a130765b8dd8498a209ea0859fc048dde06/README.md#application-objects) says that Application object can aggregate different kinds of components including CRD, but I couldn't make it work with any kind of custom resources. My humble attempt to debug controller code showed that there is no CRD resources in schema known types here:
https://github.com/kubernetes-sigs/application/blob/13e611d2b996a79c56f239d63c2679857d681aad/pkg/apis/app/v1beta1/application.go#L76-L86
Can anyone share an example Application object with CRD as component kind? Extra thanks if it would be Knative [serving](https://github.com/knative/serving) kind

janetkuo:
@barney-s @nan-yu

barney-s:
It should be supported. We will investigate and fix it.

tzununbekov:
@barney-s any updates? could you at least confirm that this issue exist?

barney-s:
@tzununbekov - We were in the midst of rearchitecting the controller. Its done now. 
Would you retest with the latest controller. Please attach the spec you used.

barney-s:
@tzununbekov - With the latest changes the we use restMapper to get the GVK for the given GK
https://github.com/kubernetes-sigs/application/blob/master/controllers/application_controller.go#L117

Could you check if this works for the Knative types.

We will add e2e test cases to cover these scenarios.

tzununbekov:
@barney-s thanks for the update, but I'm having issues with CRD deployment. Go build fails with a following error:
```
# k8s.io/client-go/rest
../../../../pkg/mod/k8s.io/client-go@v11.0.0+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher
	have (*versioned.Decoder)
	want (watch.Decoder, watch.Reporter)
```
looks like dependencies issue :thinking: 

I wanted to go with `make deploy` approach as readme suggests, but I don't like that it tries to install different tools which I already have installed (not in /bin/* as Makefile expects, but in go binaries)

Last time when I deployed Application CRD I had several local fixes in installation scripts and configs which I reset to pull latest commits. If anyone else could test described case with any CRD component kinds that would be great!

barney-s:
Regarding the build that seems odd. We do not check for compiler version may be we need to ?
[...]

PR #172, opened 2020-02-21
Use kustomize for crd based tests

barney-s:
- rename test_app to withcrd
- use kustomize layout for working and not working cases

add tests for #136


ant31:
I would not overload the docs/examples with tests cases and keep this directory clean and focused on the 'first-timer' experience/tutorial. 
Could we have add those files in tests/resources or any other places ?

barney-s:
@ant31 Moved the test files under e2e/resoruces. 
Please re-review

/assign @ant31 

FYI @nan-yu

PR #174, opened 2020-02-21
Add rbac generation rules to access other resources

ant31:
As a follow-up of #155, the file is now generated and annotations should have been added. 

cc @nan-yu


nan-yu:
@ant31 Thank you for fixing it!

PR #178, opened 2020-02-24
Add  kustomize config for prometheus ServiceMonitor

ant31:
ServiceMonitor has a Label selector that wasn't updated after kustomize transformations.

### DIFF
--- a/config/prometheus/kustomizeconfig.yaml
+++ b/config/prometheus/kustomizeconfig.yaml
@@ -1,0 +1,10 @@
+# Copyright 2020 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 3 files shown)


config/prometheus/kustomizeconfig.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2020 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+# the following config is for teaching kustomize where to look at when substituting vars.
+# It requires kustomize v2.1.0 or newer to work properly.
+
+commonLabels:
+- path: spec/selector
+  create: true
+  kind: ServiceMonitor
```
barney-s: Would we need a similar one for .metadata.labels for ServiceMonitor ?

PR #180, opened 2020-02-24
Aggregate errors and set application condition to not-ready

nan-yu:


nan-yu:
This PR is based on #167 and needs rebase when #167 is merged.

controllers/application_controller.go
```diff
@@ -58,8 +59,9 @@ func (r *ApplicationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		return ctrl.Result{}, nil
 	}
 
-	resources, err := r.updateComponents(ctx, &app)
-	newApplicationStatus := r.getNewApplicationStatus(ctx, &app, resources, err)
+	var errs []error
+	resources := r.updateComponents(ctx, &app, &errs)
```
barney-s: suggest getting errs list from updateComponents instead of passing it and updating it.
the var err []error can be defined inside the function and returned here.

controllers/application_controller.go
```diff
@@ -70,47 +72,46 @@ func (r *ApplicationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 	if app.Spec.AddOwnerRef {
 		ownerRef := metav1.NewControllerRef(app, appv1beta1.GroupVersion.WithKind("Application"))
 		*ownerRef.Controller = false
 		if err := r.setOwnerRefForResources(ctx, *ownerRef, resources); err != nil {
-			return resources, err
+			*errs = append(*errs, err)
 		}
 	}
-	return resources, nil
+	return resources
```
barney-s: return errors here

Issue #183, opened 2020-03-06
is it desired to mark Ready even if errors happened?


yuxiangqian:
introduced in:
https://github.com/kubernetes-sigs/application/pull/180

At the following line, it seems the controller will ONLY mark the application to be "ComponentsReady" when counts ready == expected AND there is some error? Is it desired?  
https://github.com/kubernetes-sigs/application/blob/master/controllers/application_controller.go#L99

nan-yu:
@yuxiangqian Thanks for catching this! The whole point of #180 is to **NOT** mark aggregated condition ready when errors happen. It is a dummy typo.
We will consider adding more test cases with mock errors.

PR #184, opened 2020-03-06
Mark aggregated condition ready only when No error happened

nan-yu:
Fixes https://github.com/kubernetes-sigs/application/issues/183

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -96,10 +96,12 @@ func (r *ApplicationReconciler) getNewApplicationStatus(ctx context.Context, app
 		Objects: objectStatuses,
 	}
 	newApplicationStatus.ComponentsReady = fmt.Sprintf("%d/%d", countReady, len(objectStatuses))
-	if aggReady && errs != nil {
+	if errs != nil {
+		setReadyUnknownCondition(newApplicationStatus, "ComponentsReadyUnknown", "failed to aggregate all components' statuses, check the Error condition for details")
+	} else if aggReady {
 		setReadyCondition(newApplicationStatus, "ComponentsReady", "all components ready")
 	} else {
-		setNotReadyCondition(newApplicationStatus, "ComponentsNotReady", "some components not ready")
+		setNotReadyCondition(newApplicationStatus, "ComponentsNotReady", fmt.Sprintf("%d components not ready", len(objectStatuses) - countReady))
 	}
 
 	if errs != nil {
--- a/controllers/condition.go
+++ b/controllers/condition.go
@@ -10,33 +10,30 @@ import (
 )
 
 func setReadyCondition(appStatus *appv1beta1.ApplicationStatus, reason, message string) {
-	setCondition(appStatus, appv1beta1.Ready, reason, message)
+	setCondition(appStatus, appv1beta1.Ready, corev1.ConditionTrue, reason, message)
 }
 
 // NotReady - shortcut to set ready condition to false
 func setNotReadyCondition(appStatus *appv1beta1.ApplicationStatus, reason, message string) {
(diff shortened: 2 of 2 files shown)


controllers/application_controller.go
```diff
@@ -96,7 +96,7 @@ func (r *ApplicationReconciler) getNewApplicationStatus(ctx context.Context, app
 		Objects: objectStatuses,
 	}
 	newApplicationStatus.ComponentsReady = fmt.Sprintf("%d/%d", countReady, len(objectStatuses))
-	if aggReady && errs != nil {
+	if aggReady && errs == nil {
 		setReadyCondition(newApplicationStatus, "ComponentsReady", "all components ready")
 	} else {
 		setNotReadyCondition(newApplicationStatus, "ComponentsNotReady", "some components not ready")
```
ant31: is it possible that all components are ready but there's an error ?
I assume yes, the message / status should reflect that or it's still confusing. 

The status would be in error with "some components not ready", as a user I would check which components isn't ready but won't find any.... 
Could we add a separate case:
something like :
```
if errs != nil { 
  "there's an error"
} else if aggReady {
 "all components ready"
} else 
{
 "some components not ready" // 'some' could be replaced with (len(objectStatuses)-countReady)
}
```

nan-yu: There are two types of conditions: `Ready` and `Error` and we need to set both.
I think we can leverage the `Unknown` status for the `Ready` condition:
```
if errs != nil {
   //set ready condition to unknown
} else if aggReady {
  "all components ready"
} else {
    "x components not ready"
}

if errs != nil {
   setErrorCondition
} else {
   clearErrorCondition
}
```

nan-yu:
Example when errors happened in aggregating statuses:
```
"status": {
    "observedGeneration": 1,
    "conditions": [
      {
        "type": "Ready",
        "status": "Unknown",
        "reason": "ComponentsReadyUnknown",
        "message": "failed to aggregate all components' statuses, check the Error condition for details",
        "lastUpdateTime": "2020-03-06T20:20:29Z",
        "lastTransitionTime": "2020-03-06T20:20:29Z"
      },
      {
        "type": "Error",
        "status": "True",
        "reason": "ErrorSeen",
        "message": "[fetchComponentListResources error, setOwnerRefForResources error, service status error, statefulset status error]",
        "lastUpdateTime": "2020-03-06T20:20:29Z",
        "lastTransitionTime": "2020-03-06T20:20:29Z"
      }
    ],
    "components": [
      {
        "link": "/api/v1/namespaces/default/services/wordpress-mysql-hsvc",
        "name": "wordpress-mysql-hsvc",
        "kind": "Service",
        "status": "Unknown"
      },
      {
        "link": "/api/v1/namespaces/default/services/wordpress-webserver-hsvc",
        "name": "wordpress-webserver-hsvc",
        "kind": "Service",
        "status": "Unknown"
      },
      {
        "link": "/api/v1/namespaces/default/services/wordpress-webserver-svc",
        "name": "wordpress-webserver-svc",
        "kind": "Service",
        "status": "Unknown"
      },
      {
        "link": "/apis/apps/v1/namespaces/default/statefulsets/wordpress-mysql",
        "name": "wordpress-mysql",
        "kind": "StatefulSet",
        "group": "apps",
        "status": "Unknown"
      },
      {
        "link": "/apis/apps/v1/namespaces/default/statefulsets/wordpress-webserver",
        "name": "wordpress-webserver",
        "kind": "StatefulSet",
        "group": "apps",
        "status": "Unknown"
      }
    ],
    "componentsReady": "0/5"
  }
```

ant31:
thank you,
/lgtm
/approve

PR #185, opened 2020-03-09
Changed service group to core

gyliu513:


nan-yu:
`core` would not work as the group of service is `v1`, same as configMap.

gyliu513:
Ah, I see, thanks @nan-yu

PR #189, opened 2020-04-13
Bump release tag to 0.8.2

barney-s:


nan-yu:
@barney-s Version `v0.8.2` was released on April 11, but the VERSION file was updated on April 12. I thought we were using the VERSION file to make the release ([VERSION_FILE=VERSION make release-branch](https://github.com/kubernetes-sigs/application/blob/master/docs/release.md#create-release-branch)). If the VERSION file wasn't updated before release, the new version should be `v0.8.1`, instead of `v0.8.2`.
Was `v0.8.2` released following the normal process?

I'm asking as it caused issues when the VERSION file was updated after release tag was created. The Louhi flow detects the release branch (v0.8.2) and uses the last commit on the branch to create the docker image. The VERSION file in the last commit points to version `v0.8.1`, which results in version inconsistency between release tag and image tag. Right now, we have to manually update the Louhi-created release branch.
I would like to make sure it is just a one-time hack, instead of a long-term thing.

PR #191, opened 2020-04-22
Add a e2e test case for backward compatibility test

nan-yu:
The test is to guarantee that existing application CR objects also work with new Application CRD and controller.
The old_app_crd.yaml was from version `v0.8.0`, which was released before we migrated to kubebuild 2.x.

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -56,6 +56,7 @@ func getKubeClientOrDie(config *rest.Config, s *runtime.Scheme) client.Client {
 
 const (
 	crdPath         = "../config/crd/bases/app.k8s.io_applications.yaml"
+	crdv080Path     = "resources/app-crd-v0.8.0.yaml"
 	testCrdPath     = "resources/withcrd/test_crd.yaml"
 	applicationPath = "resources/withcrd/base/application.yaml"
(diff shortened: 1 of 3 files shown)


e2e/resources/old-app-crd.yaml
```diff
@@ -0,0 +1,241 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
barney-s, line 1: Would you mind renaming this file to app-crd-v0.8.0 or something like that ?

e2e/main_test.go
```diff
@@ -56,6 +56,7 @@ func getKubeClientOrDie(config *rest.Config, s *runtime.Scheme) client.Client {
 
 const (
 	crdPath         = "../config/crd/bases/app.k8s.io_applications.yaml"
+	oldCrdPath      = "resources/old-app-crd.yaml"
```
barney-s, line 59: use explicit version in variable name ?

e2e/resources/old-app-crd.yaml
nan-yu: Thanks. Done

e2e/main_test.go
nan-yu: DONE

PR #193, opened 2020-06-08
Fix the controller crashing issue

nan-yu:
If no spec.selector.matchingLabels is set, the application controller
crashes due to a nil pointer. This commit fixes the issue by checking if
selector is nil before getting the matching labels from it. If no
selector is set, it matches all objects.

### DIFF
--- a/controllers/application_controller.go
+++ b/controllers/application_controller.go
@@ -116,6 +116,12 @@ func (r *ApplicationReconciler) getNewApplicationStatus(ctx context.Context, app
 func (r *ApplicationReconciler) fetchComponentListResources(ctx context.Context, groupKinds []metav1.GroupKind, selector *metav1.LabelSelector, namespace string, errs *[]error) []*unstructured.Unstructured {
 	logger := getLoggerOrDie(ctx)
 	var resources []*unstructured.Unstructured
+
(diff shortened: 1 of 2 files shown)


controllers/application_controller.go
```diff
@@ -128,7 +128,11 @@ func (r *ApplicationReconciler) fetchComponentListResources(ctx context.Context,
 
 		list := &unstructured.UnstructuredList{}
 		list.SetGroupVersionKind(mapping.GroupVersionKind)
-		if err = r.Client.List(ctx, list, client.InNamespace(namespace), client.MatchingLabels(selector.MatchLabels)); err != nil {
+		var listOptions = client.MatchingLabels{}
```
janetkuo, line 131: Is `selector.MatchExpressions` not supported?

nan-yu: It is not supported in the current version. We can open a new issue to add support for it.

nan-yu:
@janetkuo The PR is updated. Per the API doc, if `spec.selector` is nil, it matches no objects. If `spec.selector` is empty, it matches all objects.
Unit tests are added to reflect the changes. Please re-review this PR. Thanks.

PR #197, opened 2020-06-19
Fix for example kustomize secretGenerator from envs to env

cgrant:
The values for secretGenereator in the kuztomization.yaml seemed to need a String vs array. Updated it to use   env: path   vs   envs: -path


cgrant:
signed cla

PR #211, opened 2020-10-19
Improve service status

bogdanpetrea:
The status was only checked for the `LoadBalancer` type and it was also wrongly expected that `service.Spec.ClusterIP` must be empty.


bogdanpetrea:
Hi @kow3ns and @pwittrock. When might you have a look at this PR?

PR #217, opened 2021-06-21
Update to Kubernetes 1.21 and add version to ApplicationSpec.componentGVKs

nan-yu:


config/crd/bases/app.k8s.io_applications.yaml
```diff
@@ -72,23 +72,26 @@ spec:
...
                     group:
                       type: string
                     kind:
                       type: string
+                    version:
+                      type: string
                   required:
                   - group
                   - kind
+                  - version
```
janetkuo, line 94: I think we can't make a new field (i.e. `version`) required without bumping the api version, otherwise existing Application CRs will be invalid.

api/v1beta1/application_types.go
```diff
@@ -27,6 +27,14 @@ const (
 	ReasonInit = "Init"
 )
 
+// Constants for status
```
janetkuo, line 30: Why do we need this change?

nan-yu: Because KALM wants to reference to the constants. The application golang library only exposes the `api` package, but not the `pkg` package.

api/v1beta1/application_types.go
```diff
@@ -63,9 +71,9 @@ type Descriptor struct {
 
 // ApplicationSpec defines the specification for an Application.
 type ApplicationSpec struct {
-	// ComponentGroupKinds is a list of Kinds for Application's components (e.g. Deployments, Pods, Services, CRDs). It
+	// ComponentGVKs is a list of GVKs for Application's components (e.g. Deployments, Pods, Services, CRDs). It
 	// can be used in conjunction with the Application's Selector to list or watch the Applications components.
-	ComponentGroupKinds []metav1.GroupKind `json:"componentKinds,omitempty"`
+	ComponentGVKs []metav1.GroupVersionKind `json:"componentGVKs,omitempty"`
```
liggitt, line 76: isn't this a breaking change?

config/crd/bases/app.k8s.io_applications.yaml
nan-yu: I'll make the field optional.

api/v1beta1/application_types.go
liggitt: also, "GVK" doesn't seem like a great external-facing API field name

config/crd/bases/app.k8s.io_applications.yaml
```diff
@@ -72,23 +72,26 @@ spec:
                 description: AssemblyPhase represents the current phase of the application's
                   assembly. An empty value is equivalent to "Succeeded".
                 type: string
-              componentKinds:
-                description: ComponentGroupKinds is a list of Kinds for Application's
-                  components (e.g. Deployments, Pods, Services, CRDs). It can be used
-                  in conjunction with the Application's Selector to list or watch
-                  the Applications components.
+              componentGVKs:
```
liggitt, line 75: yeah, this is definitely breaking for existing data/clients

config/crd/bases/app.k8s.io_applications.yaml
nan-yu: I can manually update the yaml file to make the `version` field optional, but when I run `make manifests` to regenerate the CRD, the `version` field becomes required again. That is because Version is required in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/group_version.go#L86

Any better way to force it optional?

config/crd/bases/app.k8s.io_applications.yaml
```diff
@@ -1,12 +1,12 @@
-# Copyright 2020 The Kubernetes Authors.
+# Copyright 2021 The Kubernetes Authors.
 # SPDX-License-Identifier: Apache-2.0
 
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/application/pull/2
```
liggitt, line 8: field additions to k8.up API need to get API review, and should probably update this to point to this PR. cc @thockin

config/crd/bases/app.k8s.io_applications.yaml
nan-yu: Currently, KALM uses controller-runtime v0.19.0, which can infer the component's version from RESTMapping, but when we update controller-runtime to the latest version, the `version` field is required for an object: https://github.com/kubernetes-sigs/controller-runtime/pull/812. If we don't provide the version, we will get an error `unstructured object has no version`.

api/v1beta1/application_types.go
janetkuo: This is required for using controller-runtime v0.6.0, and we need to upgrade controller-runtime for K8s 1.21.

config/crd/bases/app.k8s.io_applications.yaml
nan-yu: What about we keep the original name `componentKinds`, but add an optional `version` field? 
This is tricky because if we re-generate the manifests, the version field will become required again.

api/v1beta1/application_types.go
liggitt: > The version field is required for using controller-runtime v0.6.0+ (because of [kubernetes-sigs/controller-runtime#812](https://github.com/kubernetes-sigs/controller-runtime/pull/812)), and we need to upgrade controller-runtime for using v1 CRDs.

including version in the reference is not recommended... it means that any change in the available versions of the referenced types (e.g. apps/v1beta1 being removed in 1.16 in favor of apps/v1) would invalidate/require sweeping all references and updating them (e.g. to apps/v1). It would be better to resolve the preferred version of the referenced group/kind expressed in this API in the code that uses it and constructs the controller-runtime client

liggitt: xref https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#generic-object-reference for reference guidance

config/crd/bases/app.k8s.io_applications.yaml
liggitt: If version inference is no longer built into controller-runtime, hoisting the version inference out into your own code would be better than baking it into the version reference here. See comment at https://github.com/kubernetes-sigs/application/pull/217/files#r655813175

PR #228, opened 2022-12-14
when ownerRef is same, needless to Update resource

xiaoping378:
needless to update the same onwerRef when Sync every sync-period


xiaoping378:
@barney-s @kow3ns any points ? please review

